Issue #30, opened 2015-12-02
Issues with Content-assist


tfesenko:
## Content-assist for patternProperties

Invoke content-assist for response code:
<img width="380" alt="screen shot 2016-01-27 at 5 29 02 pm" src="https://cloud.githubusercontent.com/assets/644582/12630671/86491238-c51b-11e5-991a-aaa761c9f4f2.png">

The first option is `([0-9]{3})$|^(default)$`, choosing it will create an invalid model:

``` yaml
  /taxFilings:
    get:
      responses:
        ([0-9]{3})$|^(default)$:
```

PatternProperties probably are probably more useful in proposal description than for replacementString.

tfesenko:
## `path` value for method parameters `in`

Line 26 of the default model defines the `"path"` value. However, content-assist does not provide it.

``` yaml
        in: "path"
```

tfesenko:
## "$ref" key not provided by content-assist

UPD: now it works

Line 34 of default model:

``` yaml
          schema:
            $ref: "#/definitions/TaxFilingObject"
```

tfesenko:
## `type` keyword in definitons and properties

There is no code assist for type value of definition or property, e.g. 

``` yaml
definitions:
  TaxFilingObject:
    type: "object"
```

and (for `filingID`)

``` yaml
  TaxFilingObject:
    type: "object"
    description: "An individual Tax Filing record, accessed by its ID"
    properties:
      filingID:
        type: "string"
```

Parameters type works fine.

tfesenko:
## Content-assist for value of `schema/$ref`

It would be good to provide content assist for `responses/schema/$ref` which would calculate available refs to definitions.

ghillairet:
### `path value for method parameters in` issue

It is solved in this pull request:
https://github.com/ModelSolv/SwagEdit/pull/34

### `Content-assist for patternProperties`

This raises the question about how we should handle pattern properties, and if we should introduce specific implementation for some of them. Some pattern properties are fine to be use as replacement strings, see below for example on path, some are not like for responses code, that uses regex strings.

![screen shot 2015-12-04 at 17 49 04](https://cloud.githubusercontent.com/assets/148045/11595387/9bdb8004-9aaf-11e5-9553-a7c48a9918d5.png)

### refs issues

The `refs` issues are a bit different and involves two different cases:
[...]

PR #34, opened 2015-12-04
[#30] Issues with Content-assist

ghillairet:
This solves the issue with the content assist proposal for parameters `in`.


tfesenko:
Content assist for the value of `parameters`/`in` works good now. 
But a side effect of this fix are rethought values for boolean:
screen shot 2015-12-04 at 3 10 39 pm

com.reprezen.swagedit/src/com/reprezen/swagedit/validation/SwaggerSchema.java
```diff
@@ -94,80 +97,110 @@ else if (node.has("properties")) {
...
+		final String[] paths = path.split(":");
+		Set<JsonNode> definitions = Collections.emptySet();
+		Set<JsonNode> toTraverse = Collections.singleton(asJson());
 
-		for (String current : paths) {
-			JsonNode next = getDefinition(definition, current);
-			if (next != null) {
-				definition = next;
+		for (String currentPath: paths) {
+			definitions = new HashSet<>();
```
tfesenko, line 120: It looks strange that we reset `definitions` so the overall result of `getDefinitionForPath()` will be actually the value of the last pass only. First I thought it was a bug. It takes some time to figure out that this is the intended behavior. What about a little commit just to explain it?

com.reprezen.swagedit/src/com/reprezen/swagedit/validation/SwaggerSchema.java
```diff
@@ -94,80 +97,110 @@ else if (node.has("properties")) {
...
 	/**
-	 * Returns the schema definition that corresponds to the given path.
+	 * Returns the set of definitions that matches the given path.
 	 * 
 	 * @param path
-	 * @return schema definition
+	 * @return set of definitions for the path
 	 */
-	public JsonNode getDefinitionForPath(String path) {
+	public Set<JsonNode> getDefinitionForPath(String path) {
```
tfesenko: What about making it plural - `getDefinition*s*ForPath()`?

tfesenko:
Looks good. Thanks!

Issue #32, opened 2015-12-02
Incorrect validation marker location


tfesenko:
Take this model and try to find what's wrong with it.
The validation marker points to line 14:
<img width="457" alt="screen shot 2015-12-02 at 6 03 57 pm" src="https://cloud.githubusercontent.com/assets/644582/11547054/1ac74ba6-991f-11e5-81e9-5bacfd50f5fb.png">

However, the error (the only error) is on line 33 - `schema:` which missing the `$ref`.

Model:

``` yaml
swagger: "2.0"
info:
  description: "Tax Blaster"
  version: "1.0.0"
  title: "TaxBlaster"
host: "taxblaster.com"
basePath: "/api"
tags:
- name: "TaxFilingObject"
  description: "An individual Tax Filing record, accessed by its ID"
schemes:
- "http"
paths:
  /taxFilings/{id}:
    get:
      tags:
      - "TaxFilingObject"
      description: ""
      operationId: "getTaxFiling"
      consumes:
      - "application/json"
      produces:
      - "application/json"
      parameters:
      - name: "id"
        in: "path"
        description: "filingID of the requested TaxFiling"
        required: true
        type: "string"
      responses:
        200:
          description: ""
          schema:
        404:
          description: ""
    parameters:
    - name: "id"
      in: "path"
      description: "filingID of the requested TaxFiling"
      required: true
      type: "string"

definitions:
  TaxFilingObject:
    type: "object"
    description: "An individual Tax Filing record, accessed by its ID"
    properties:
      filingID:
        type: "string"
      jurisdiction:
        type: "string"
      year:
        type: "string"
      period:
        type: "integer"
      currency:
        type: "string"
      grossIncome:
        type: "number"
      taxLiability:
        type: "number"
      taxpayer:
        type: "object"
```

tfesenko:
Similar with this model:

``` yaml
swagger: "2.0"
info:
  description: "Tax Blaster"
  version: "1.0.0"
  title: "TaxBlaster"
host: "taxblaster.com"
basePath: "/api"
tags:
- name: "TaxFilingObject"
  description: "An individual Tax Filing record, accessed by its ID"
schemes:
- "http"
paths:
  /taxFilings/{id}:
    get:
      tags:
      - "TaxFilingObject"
      description: ""
      operationId: "getTaxFiling"
      consumes:
      - "application/json"
      produces:
      - "application/json"
      parameters:
      - name: "id"
        in: "path"
[...]

PR #35, opened 2015-12-04
[#32] Incorrent validation marker location

ghillairet:
This will gives better error results, this is done by collecting all validation reports from the json schema validator.

### DIFF
--- a/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/ValidatorTest.xtend
+++ b/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/ValidatorTest.xtend
@@ -7,6 +7,7 @@ import org.eclipse.core.resources.IMarker
 import org.junit.Test
 
 import static org.junit.Assert.assertEquals
+import static org.junit.Assert.assertTrue
 
 class ValidatorTest {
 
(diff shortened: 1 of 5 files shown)


com.reprezen.swagedit/src/com/reprezen/swagedit/validation/Validator.java
```diff
@@ -33,17 +35,17 @@
 	private static final SwaggerSchema schema = new SwaggerSchema();
 
 	/**
-	 * Returns a list or errors if validation fails. 
+	 * Returns a list or errors if validation fails.
```
tfesenko, line 38: It now returns a `Set`, but a javadoc says it's list. Why don't we replace it with a `Collection`?

com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/ValidatorTest.xtend
```diff
@@ -149,4 +150,31 @@ class ValidatorTest {
...
+		          schema:
+		'''
+
+		document.set(content)
+		val errors = validator.validate(document)
+
+		assertEquals(5, errors.size())
+
+		errors.forEach[
+			assertTrue(it.line == 10 || it.line == 11)
```
tfesenko, line 176: You can also use a range: `(10..11).contains(it.line)`

tfesenko:
Thanks! PR passed code review and QA.
Validation is so much better with this fix.

PR #38, opened 2015-12-10
Feature/live validation

ghillairet:
Support activation of validation when user make changes to the document.


com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerDocument.java
```diff
@@ -279,4 +263,21 @@ private String getId(Node node) {
 		}
 	}
 
+	public void onChange() {
```
tfesenko, line 266: I would name it `handleChange()`, but it's a matter of taste

com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerEditor.java
```diff
@@ -53,6 +77,13 @@ protected YEditSourceViewerConfiguration createSourceViewerConfiguration() {
 		configuration.setEditor(this);
 		return configuration;
 	}
+
+	@Override
+	protected void doSetInput(IEditorInput input) throws CoreException {
+		super.doSetInput(input);
+		
+		getDocumentProvider().getDocument(getEditorInput()).addDocumentListener(changeListener);
```
tfesenko, line 85: The `changeListener` is added, but never removed.

tfesenko:
Looks good.
@ghillairet , can you please check if the `changeListener` should be removed from a document?

Issue #41, opened 2015-12-17
Code assist should not put unnecessary quotes around string value


tedepstein:
I tried Ctrl+space after the `in` key at the end of this code snippet, and it immediately put in the quoted value `"body"`.  We should only use quotes where they are required.

```
paths:
  /products:
    get:
      summary: Product Types
      description: |
        The Products endpoint returns information about the *Uber* products
        offered at a given location. The response includes the display name
        and other details about each product, and lists the products in the
        proper display order.
      parameters:
        - name: latitude
          in: "body"
```

tfesenko:
Still present in the product.

ghillairet:
Double quotes have been removed from those proposals.

Also changed how the string proposals behavior, when the proposal is a string, it will not add empty double quotes when the user insert the proposal, and the display has been changed to this:

![screen shot 2016-01-08 at 16 38 40](https://cloud.githubusercontent.com/assets/148045/12201855/c968f2c6-b626-11e5-920b-1a27f16f5120.png)

Also added an empty proposal for when no solutions can be found.

![screen shot 2016-01-08 at 16 38 33](https://cloud.githubusercontent.com/assets/148045/12201865/d81b0f7a-b626-11e5-9a34-c65b007f8b41.png)

tedepstein:
In progress.  Test failure and some issues noted in the pull request #63

tedepstein:
@tfesenko , Guillaume has added a new commit last week.  Please have a look.

tfesenko:
@tedepstein , I only see commit 8c12723 which was merged to the master as a part of PR #41 .
It works. Why the issue was not closed then? Probably I just forgot to close the issue.

PR #63, opened 2016-01-08
Feature/41

ghillairet:
See #41


com.reprezen.swagedit/src/com/reprezen/swagedit/assist/EmptyCompletionProposal.java
```diff
@@ -0,0 +1,39 @@
...
+		return null;
+	}
+
+	@Override
+	public String getAdditionalProposalInfo() {
+		return null;
+	}
+
+	@Override
+	public String getDisplayString() {
```
tfesenko, line 25: This class looks just like `NoCompletionProposal`. It takes several minutes to figure out that the difference is in `getDisplayString()`. Can we reuse the code?

tfesenko, line 25: The order of methods is different, which makes it difficult to compare by file diff. You have to actually read both classes to find the real diff.

com.reprezen.swagedit/src/com/reprezen/swagedit/assist/SwaggerContentAssistProcessor.java
```diff
@@ -77,12 +79,37 @@ public SwaggerContentAssistProcessor() {}
...
 
-		final ICompletionProposal[] result = new ICompletionProposal[proposals.size()];
-		for (int i = 0; i < proposals.size(); i++) {
-			result[i] = proposals.get(i);
+		if (proposals.isEmpty()) {
+			proposals.add(new NoCompletionProposal());
+		}
+
+		// make sure the proposal dialog shows up
+		if (proposals.size() == 1) {
```
tfesenko, line 87: Why a single proposal will not show up? Is it a bug? I believe it's useful, but I just really don't understand the need in this check - `proposals.size() == 1`.

tfesenko:
There is a test failure on this branch:

```
SwaggerContentAssistProcessorTest.shouldProvideEndOfWord:69 expected:<1> but was:<2>
```

tedepstein:
Passed QA Code review, merging.

com.reprezen.swagedit/src/com/reprezen/swagedit/assist/SwaggerContentAssistProcessor.java
ghillairet, line 87: I removed those, this could be done by options to set in the ContentAssistant.

ghillairet:
See new PR #78

Issue #68, opened 2016-01-13
Provide explanatory comments and example message in default Swagger spec


tedepstein:
When creating a new Swagger spec, SwagEdit includes some default/boilerplate content.  We should include some comments to help new users understand Swagger, and to help with visual orientation (by introducing some variation in spacing and color).

We should also introduce a message example, so it works right away in editors that support example, driven mock services.  (Like our own RepreZen API Studio).

Here's a revised default model:

```

---
# This is a sample Swagger spec, describing a simple API as a starting point.  
swagger: "2.0"
info:
  description: "Tax Blaster"
  version: "1.0.0"
  title: "TaxBlaster"
host: "taxblaster.com"
basePath: "/api"

# Tags organize operations into groups for presentation in the Swagger UI.
# Each tag has an optional description, which the Swagger UI will display in 
# the tag group header.
tags:
- name: "TaxFilingObject"
  description: "An individual Tax Filing record, accessed by its ID"
schemes:
- "http"

paths:

  # Each Path Item Object describes a resource, containing a set of operations
  # at a specified path.  The Path Item object can define parameters and 
  # responses common all of its contained operations.
  /taxFilings/{id}:
    parameters:
    - name: "id"
      in: "path"
      description: "filingID of the requested TaxFiling"
      required: true
      type: "string"

    # Operations are identified by an HTTP method.  
    get:
      tags:
      - "TaxFilingObject"
      description: ""
      operationId: "getTaxFiling"
      consumes:
      - "application/json"
      produces:
      - "application/json"
      parameters:
      - name: "id"
        in: "path"
        description: "filingID of the requested TaxFiling"
        required: true
        type: "string"
      responses:
        200:
          description: ""
          schema:
            # Reference to a Schema Object described in 'definitions' section
            $ref: "#/definitions/TaxFilingObject"
          examples:
            # Example message, keyed by media type
            application/json : 
              {
                "filingID" : "1234",
                "jurisdiction" : "Federal",
                "year" : "2015",
                "period" : "CALENDAR_YEAR",
                "currency" : "EUR",
                "grossIncome" : "74832",
                "taxLiability" : "15640"
              }
        404:
          description: ""

# The definitions section contains a set of named Schema Objects.  Each schema
# object describes a reusable data type, which can be reference by name.
definitions:
  TaxFilingObject:
    type: "object"
    description: "An individual Tax Filing record, accessed by its ID"
    properties:
      filingID:
        type: "string"
[...]

PR #76, opened 2016-01-14
[#68] Provide explanatory comments and example message in default Swa…

ghillairet:
…gger spec

Updated the default swagger spec.


tfesenko:
Looks good

Issue #44, opened 2015-12-17
Beautify the default color scheme


tedepstein:
The default color scheme is not as pretty as it might be.  I'll play with this a bit and make some more specific suggestions.

tedepstein:
In API Studio, we are using the Eclipse Color Theme plugin, with a dark color theme (inkpot) selected.  For general use outside of API studio, we probably should not assume a dark background, so we should choose colors that works with a light background.

ghillairet:
Note that eclipse color theme plugins requires that mappings for specific editors are contributed (see https://github.com/eclipse-color-theme/eclipse-color-theme/wiki/Contributing). The mapping for YEdit is already present in the current version of eclipsecolortheme, that explains why SwagEdit works well with it.

Also the general use of this color theme plugin is that users choose a theme and then all their editors use that particular theme.

The default theme from Eclipse is the one with white background and purple color for keywords (see Java editor for ex). Maybe this one should be use.

tedepstein:
@ghillairet , can you paste a screenshot with SwagEdit in the default theme?

ghillairet:
So the default theme for YEdit does not use similar colors as the Eclipse default. Choosing Default theme in eclipse color themes preferences makes editors use their defaults preferences.

Looks like this for YEdit (and SwagEdit)

![screen shot 2016-01-07 at 16 44 32](https://cloud.githubusercontent.com/assets/148045/12174756/c0d640a0-b55e-11e5-9d73-07c2bf3a452c.png)

I was thinking we could provide default preferences that would look like that

![screen shot 2016-01-07 at 16 46 03](https://cloud.githubusercontent.com/assets/148045/12174763/d009aa6c-b55e-11e5-872d-2636b3daa2c6.png)

tedepstein:
Yes, I think that second screenshot looks good.  Let's go with that color scheme as the default.

tfesenko:
After having merged the PR, I noticed that what I see is slightly different from the screenshot above:
<img width="703" alt="screen shot 2016-01-20 at 4 18 50 pm" src="https://cloud.githubusercontent.com/assets/644582/12463530/f4649506-bf91-11e5-9db6-48e3617db073.png">

Note that boolean values ("Constants") are blue in my screenshot, but they are green on Guillaume's.

tedepstein:
[...]

PR #81, opened 2016-01-15
[#44] Beautify the default color scheme

ghillairet:
This is the pull request to add a default theme for SwagEdit, see #44 

It involves some changes since to have default colors for SwagEdit, we need to use a different preference store than the one from YEdit.

Also since we want eclipse-color-theme to work with SwagEdit, and we use different preferences from YEdit, we need to provide a SwagEdit preference mapping file for eclipse-color-theme (this is the new xml file in folder resources). This also includes a new dependency to eclipse-color-theme, but I marked it as optional. Without that mapping file, the colors of the SwagEdit editor would not be changed if the user chooses a new color theme, only the background would change.

### DIFF
--- a/com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerScanner.java
+++ b/com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerScanner.java
@@ -1,0 +1,175 @@
+package com.reprezen.swagedit.editor;
+
+import java.util.ArrayList;
+
+import org.dadacoalition.yedit.editor.ColorManager;
+import org.dadacoalition.yedit.editor.scanner.AnchorWordDetector;
+import org.dadacoalition.yedit.editor.scanner.DocumentStartAndEndRule;
+import org.dadacoalition.yedit.editor.scanner.DoubleQuotedKeyRule;
+import org.dadacoalition.yedit.editor.scanner.IndicatorCharacterRule;
+import org.dadacoalition.yedit.editor.scanner.KeyRule;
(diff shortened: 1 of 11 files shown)


com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerScanner.java
```diff
@@ -0,0 +1,175 @@
...
+	public SwaggerScanner(ColorManager colorManager, IPreferenceStore store) {
+		super(colorManager);
+
+		this.colorManager = colorManager;
+		this.store = store;
+		init();
+	}
+
+	protected void init() {
+		TextAttribute keyAttr = tokenAttribute(
```
tfesenko, line 53: I see, we have to copy this code because the preferences store is defined as a variable in a method and super methods are private.

com.reprezen.swagedit/src/com/reprezen/swagedit/preferences/SwaggerColorPreferences.java
```diff
@@ -1,12 +1,19 @@
...
-import org.dadacoalition.yedit.Activator;
 import org.dadacoalition.yedit.preferences.ColorPreferences;
 
+import com.reprezen.swagedit.Activator;
+
+/*
+ * This implementation of preference page overrides the YEdit implementation but 
+ * uses it's own preference store.
+ * 
+ */
```
tfesenko, line 11: Thanks

tfesenko:
Passed code review

tfesenko:
QA, looks as advertised:
screen shot 2016-01-20 at 4 18 50 pm

Issue #97, opened 2016-02-11
Templates defined in the Path context show in other contexts


tedepstein:
The code templates we've provided for the Path context appear in the suggestion list in other contexts:

![image](https://cloud.githubusercontent.com/assets/692813/12986051/52b0ce30-d0c3-11e5-9dae-a75e242e4a50.png)

![image](https://cloud.githubusercontent.com/assets/692813/12986097/8e9979b0-d0c3-11e5-8a1e-8ee1b533f074.png)

There are some contexts where they do not appear, so it's not every context.  But some of them, where we should not be allowed to define a method, still show those method templates.

tedepstein:
@tfesenko , this is also high priority, as it's interfering with the video demo that @daffinm needs to record.

tedepstein:
Resolved

Issue #31, opened 2015-12-02
More code template contexts


tfesenko:
We only support code templates for `root` and `path`.
Code templates for resources, methods and responses would be helpful too.

tedepstein:
It would also be really helpful to have a context for any location where a Schema Object is allowed or expected.  Usually, this is under a `schema` key.

tedepstein:
Resolved.  Closing.

PR #101, opened 2016-02-19
#31, #97 - More Template Contexts

tfesenko:
This PR addresses the following issues:
#31 - More Template Contexts
#97 - Templates defined in the Path context show in other contexts

You can see all contexts and how they are defined in `SwaggerContextType`
Tests are in `CodeTemplateContextTest`
`SwaggerTemplateContext` provides support for indentation levels - we can use the same code template in different indentation levels.
## Code templates

Code templates are defined in `com.reprezen.swagedit/resources/templates.xml`
I provided very basic code templates for all context types just to make it easier to test this PR

Schema/allOf/[] is not supported yet because its format is different from other Schema elements

### DIFF
--- a/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/CodeTemplateContextTest.java
+++ b/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/CodeTemplateContextTest.java
@@ -30,6 +30,8 @@ public class CodeTemplateContextTest {
 		assertThat(getContextType(":paths:"), not(equalTo(PathItemContextType.CONTEXT_ID)));
 		assertThat(getContextType(":paths:/pets:get"), not(equalTo(PathItemContextType.CONTEXT_ID)));
 		assertThat(getContextType(":paths:/pets:get:responses"), not(equalTo(PathItemContextType.CONTEXT_ID)));
+		assertThat(getContextType(":paths:/my-pets/v1/{pet-id}:somethingElse"),
+				not(equalTo(PathItemContextType.CONTEXT_ID)));
 	}
 
 	@Test
@@ -70,6 +72,15 @@ public class CodeTemplateContextTest {
 				equalTo(SchemaContextType.CONTEXT_ID));
 		assertThat(getContextType(":definitions:TaxFilingObject:additionalProperties"),
 				equalTo(SchemaContextType.CONTEXT_ID));
+	}
+
+	@Test
+	public void test$InRegex() throws Exception {
+		assertFalse("abcd".matches("abc"));
+		assertTrue("abc".matches("abc"));
+
+		assertFalse("abcd".matches("abc$"));
+		assertTrue("abc".matches("abc$"));
 	}
 
 }
--- a/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/CodeTemplateContextTest.java
+++ b/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/CodeTemplateContextTest.java
@@ -24,7 +24,8 @@ public class CodeTemplateContextTest {
 		assertThat(getContentType(":paths:/my-pets/v1/{pet-id}"), equalTo(PathItemContextType.CONTEXT_ID));
 		assertThat(getContentType(":paths:/pets"), equalTo(PathItemContextType.CONTEXT_ID));
 
-		// tests for #Templates defined in the Path context show in other contexts
+		// tests for #Templates defined in the Path context show in other
+		// contexts
 		assertThat(getContentType(":paths:"), not(equalTo(PathItemContextType.CONTEXT_ID)));
 		assertThat(getContentType(":paths:/pets:get"), not(equalTo(PathItemContextType.CONTEXT_ID)));
 		assertThat(getContentType(":paths:/pets:get:responses"), not(equalTo(PathItemContextType.CONTEXT_ID)));
@@ -58,6 +59,10 @@ public class CodeTemplateContextTest {
 		assertThat(getContentType(":paths:/pets/{id}:get:responses:200:schema:"),
 				equalTo(SchemaContextType.CONTEXT_ID));
 		assertThat(getContentType(":paths:/pets:post:parameters:@0:schema:"), equalTo(SchemaContextType.CONTEXT_ID));
+		assertThat(getContentType(":paths:/pets:post:parameters:@0:schema:properties:name"),
+				equalTo(SchemaContextType.CONTEXT_ID));
+		assertThat(getContentType(":paths:/pets:post:parameters:@0:schema:items"),
+				equalTo(SchemaContextType.CONTEXT_ID));
 	}
-	
+
 }
--- a/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/CodeTemplateContextTest.java
+++ b/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/CodeTemplateContextTest.java
@@ -1,0 +1,33 @@
+package com.reprezen.swagedit.tests;
+
+import static org.junit.Assert.*;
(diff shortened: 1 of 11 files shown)


tfesenko:
@tedepstein , can you please review?

com.reprezen.swagedit/resources/templates.xml
```diff
@@ -25,21 +25,34 @@ info:
...
 
      <template name="paths" 
-       id="com.reprezen.swagedit.templates.paths_template" 
-       description="swagger template" 
-       context="com.reprezen.swagedit.templates.swagger.root" 
-       enabled="true">
+ id="com.reprezen.swagedit.templates.paths_template" 
+ description="swagger template" 
+ context="com.reprezen.swagedit.templates.swagger.root" 
+ enabled="true">
```
tedepstein: Looks like a small formatting problem.

com.reprezen.swagedit/resources/templates.xml
```diff
@@ -25,21 +25,34 @@ info:
...
           description: Ok
-     </template>
-
+</template>
+     
+     <template name="oath2" 
+       id="com.reprezen.swagedit.templates.securityDef_oath2" 
+       description="OAuth2 security def" 
+       context="com.reprezen.swagedit.templates.swagger.security_def" 
+       enabled="true">oauth2:
```
tedepstein: if this key is the name of the security scheme, maybe we should call it "OAuth2Scheme", so the name and the type are different.

com.reprezen.swagedit/resources/templates.xml
```diff
@@ -25,21 +25,34 @@ info:
...
+ type: oauth2
```
tedepstein: No indent here.  Does this mean that it goes at the current indent level?  Current indent level +1?

com.reprezen.swagedit/src/com/reprezen/swagedit/templates/SwaggerContextType.java
```diff
@@ -21,14 +26,79 @@ private void addGlobalResolvers() {
 	}
 
 	public static String getContentType(String path) {
```
tedepstein: Should `getContentType` be `getContextType` ?

com.reprezen.swagedit/src/com/reprezen/swagedit/templates/SwaggerContextType.java
```diff
@@ -21,14 +26,79 @@ private void addGlobalResolvers() {
...
+			return SecurityDefContextType.CONTEXT_ID;
+		}
+		if (path.equals(":paths")) {
+			return PathsContextType.CONTEXT_ID;
+		}
+		if (path.matches(":paths:(/[^:]*)+")) { // /paths/[pathItem]/
+			return PathItemContextType.CONTEXT_ID;
+		}
+		if (path.equals(":responses")//
+				|| path.matches(":paths:/[^:]*:[^:]*:responses")) {
```
tedepstein: These regexes are getting repeated.  Can we DRY this up by defining constants for PATH_ITEM_REGEX, METHOD_PARAMETERS_REGEX, etc.?

com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/CodeTemplateContextTest.java
```diff
@@ -0,0 +1,69 @@
...
+	@Test
+	public void testResponses() throws Exception {
+		assertThat(getContentType(":responses"), equalTo(ResponsesContextType.CONTEXT_ID));
+		assertThat(getContentType(":paths:/resource:get:responses"), equalTo(ResponsesContextType.CONTEXT_ID));
+	}
+
+	@Test
+	public void testParameters() throws Exception {
+		assertThat(getContentType(":paths:/taxFilings/{id}:get:parameters:@0:"),
+				equalTo(ParametersContextType.CONTEXT_ID));
```
tedepstein: The context shown in the regex is where we're already defining the first element of the collection.  So it's  a single specific parameter.  In this case, I think `ParametersContextType` should be the singular `ParameterContextType`.  

Small point maybe, but there are several cases like this in Swagger, where we have a `Paths Object` (collection) and a `Path Item Object`, a `Responses Object` (collection) and a `Response Object`, etc.  So for clarity and consistency, I think it's important to name these singular or plural, as appropriate.

com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/CodeTemplateContextTest.java
```diff
@@ -0,0 +1,69 @@
...
+		assertThat(getContentType(":paths:/taxFilings/{id}:get:parameters"), equalTo(ParametersContextType.CONTEXT_ID));
```
tedepstein: This one is the parameters collection.  I'm suggesting we split parameters vs. individual parameter, as two separate contexts.

com.reprezen.swagedit/src/com/reprezen/swagedit/templates/SwaggerContextType.java
```diff
@@ -21,14 +26,79 @@ private void addGlobalResolvers() {
...
+		if (/* path.equals(":parameters") || */ // is an object, not an array
+		path.matches(":parameters:[^:]*") //
+				|| path.matches(":paths:/[^:]*:[^:]*:parameters(:@\\d+)?")
+				|| path.matches(":paths:/[^:]*:parameters")) {
+			return ParametersContextType.CONTEXT_ID;
+		}
+		if (path.matches(":definitions:[^:]*") //
+				|| path.matches(".*:parameters(:@\\d+):schema")//
+				|| path.matches(".*:parameters(:@\\d+):schema:items")//
+				|| path.matches(".*:parameters(:@\\d+):schema:properties:[^:]+")//
```
tedepstein: `patternProperties` and `additionalProperties` can also be schemas.  
http://json-schema.org/latest/json-schema-validation.html#rfc.section.5.4.4

`additionalProperties` can be a boolean value, OR it can be a schema object.  I think/hope we can rely on the schema-generated keyword suggestions to include True, False, and the various schema keywords in this context.  So, from a code template perspective, we would just call this SchemaContextType.CONTEXT_ID.  No need to define a separate context for this.

`patternProperties` is tricky.  It's an object, having one "Pattern Property" for each allowed pattern.  The name of each Pattern Property is a regular expression, and the value is a JSON Schema object.  To match this correctly, we would need a regular expression that matches regular expressions.  I am fine with skipping this one, or creating a very general regex to match the Pattern Property name.

com.reprezen.swagedit/src/com/reprezen/swagedit/templates/SwaggerTemplateContext.java
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		@Override
+		public TemplateBuffer translate(Template template) throws TemplateException {
+			return translate(template.getPattern());
+		}
+
+		@Override
+		public TemplateBuffer translate(String string) throws TemplateException {
+			String withIndentation = string.replaceAll("(\r\n?)|(\n)", "$0" + indentation);
```
tedepstein, line 120: Nice!

tedepstein:
@tfesenko , code review completed, with a few comments.

tfesenko:
@tedepstein , I addressed your comments. Can you please review?

tedepstein:
@tfesenko , reviewed, with one important issue found.

tedepstein:
@tfesenko wrote:

> method parameters are covered by `([^:]+:)?` in `PARAMETERS_LIST_REGEX`, they are verified be tests in 954ccfa , they are available here...  <snip>
> Does it cover what you expected for method-level parameters?

As discussed yes.  I didn't look closely enough at that regex and those test cases.  Glad to see it's covered.  

Code review passed, merging now.

Issue #67, opened 2016-01-13
Non-Swagger YAML files should open in YEdit, not be validated as Swagger files


tedepstein:
SwagEdit should, as far as possible, co-exist peacefully with YEdit.  I found in API Studio that when I install both YEdit and SwagEdit, all YAML files open in SwagEdit by default, and are validated as Swagger files.  For example, they will fail validation because the `swagger` property is not present.

I tried in a new installation of Eclipse Mars and observed similar behavior.  There should be some way for SwagEdit to determine if a YAML file is "trying" to be a Swagger spec.  Presence/absence of the swagger attribute would be a reasonable test.  If it's not trying to be a Swagger spec, then the file should open in YEdit, not SwagEdit, and should have SwagEdit's validations, code assist and other behaviors applied to it.

tedepstein:
Flagged this one as high priority, not for the current release (API Studio 0.9.2), but for the next release (SwagEdit Open Source Release).

tedepstein:
Resolved.  Closing.

PR #103, opened 2016-02-19
#67 - Non-Swagger YAML files should open in YEdit, not be validated as Swagger files

tfesenko:


tedepstein:
Looks dead simple, passed code review.  
When I get a moment, I will test this by installing YEdit and SwagEdit side-by-side in a clean Eclipse distro.

tedepstein:
@tfesenko , merging these open pull requests, so we can integrate SwagEdit.

Issue #64, opened 2016-01-08
Add copyright headers to the sources


tfesenko:
Example: a copyright header from JDT:

```
/*******************************************************************************
 * Copyright (c) 2000, 2015 IBM Corporation and others.
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v1.0
 * which accompanies this distribution, and is available at
 * http://www.eclipse.org/legal/epl-v10.html
 *
 * Contributors:
 *     IBM Corporation - initial API and implementation
 *******************************************************************************/
```

tedepstein:
Ted to update text

tedepstein:
Here's the copyright notice we should use:

```
/*******************************************************************************
 * Copyright (c) 2016 ModelSolv, Inc. and others.
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v1.0
 * which accompanies this distribution, and is available at
 * http://www.eclipse.org/legal/epl-v10.html
 *
 * Contributors:
 *    ModelSolv, Inc. - initial API and implementation and/or initial documentation
 *******************************************************************************/
```

tedepstein:
Merged, resolved.

PR #106, opened 2016-02-29
#64 - Add copyright headers to the sources

tfesenko:


andylowry:
@tanya, Review passed. All looks good, except that I did find one source file that wasn't covered: `com.reprezen.swagedit/src/com/reprezen/swagedit/json/schema.json`. I know json doesn't generally allow comments, but the jackson parser has an option to allow ignore /\* comments */, so it might be worth covering this file as well.

tfesenko:
`com.reprezen.swagedit/src/com/reprezen/swagedit/json/schema.json` is JSON Schema for Swagger. We did not created, we took an existing file and the added minor modifications.

Issue #87, opened 2016-01-21
oneOf Schema Violation: Provide names for schemas


tedepstein:
As discussed in my comments on #60, it would be very helpful to reference candidate schemas by a descriptive name, instead of a JSON Pointer.  Names could be taken from the key in the `definitions` section, or the [`title` keyword](http://json-schema.org/latest/json-schema-validation.html#anchor98) and/or from an extended property added to the JSON schema.

tedepstein:
Also applies to `anyOf`.

tedepstein:
@tfesenko , this was addressed in #108  and #110 .  Closing, but please take a quick look & reopen if you think there is further work that still needs to be done on this.

PR #108, opened 2016-03-04
#87 Reference candidate schemas by a descriptive name

tfesenko:
@tedepstein, the new error message looks like this:
screen shot 2016-03-04 at 4 55 09 pm

I replaced the location by a human readable label and sorted the messages so the best candidate (least amount of errors) is displayed first.

### DIFF
--- a/com.reprezen.swagedit/src/com/reprezen/swagedit/validation/SwaggerError.java
+++ b/com.reprezen.swagedit/src/com/reprezen/swagedit/validation/SwaggerError.java
@@ -10,7 +10,9 @@
  *******************************************************************************/
 package com.reprezen.swagedit.validation;
 
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.Map;
 import java.util.Set;
 
@@ -18,7 +20,10 @@ import org.eclipse.core.resources.IMarker;
 import org.yaml.snakeyaml.error.MarkedYAMLException;
 import org.yaml.snakeyaml.error.YAMLException;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.base.Strings;
+import com.google.common.collect.Lists;
+import com.reprezen.swagedit.json.JsonSchemaManager;
 
 public class SwaggerError {
 
@@ -31,6 +36,7 @@ public class SwaggerError {
 	public int level;
 	public int line;
 	public int indent = 0;
+	private static final JsonNode swaggerSchema = new JsonSchemaManager().getSwaggerSchema().asJson();;
 
 	public SwaggerError(int line, int level, String message) {
 		this.line = line;
@@ -75,7 +81,7 @@ public class SwaggerError {
 
 			return builder.toString();
 		}
-		
+
 		return message;
 	}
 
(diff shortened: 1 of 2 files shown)


tedepstein:
Oh, @tfesenko , that is _so_ much better!!!

tfesenko:
@tedepstein , and no changes to the Swagger Schema! 
I can create more screenshots for you then.

com.reprezen.swagedit/src/com/reprezen/swagedit/validation/SwaggerError.java
```diff
@@ -109,20 +128,66 @@ public String getMessage() {
...
+			JsonNode title = swaggerSchemaNode.get("title");
+			if (title != null) {
+				return title.asText();
+			}
+			JsonNode description = swaggerSchemaNode.get("title");
+			if (description != null) {
+				return description.asText();
+			}
+			// "$ref":"#/definitions/headerParameterSubSchema"
+			JsonNode ref = swaggerSchemaNode.get("$ref");
```
tedepstein, line 156: We might have to be careful with Refs.  I remember seeing in a discussion that JSON Refs can use JSON Pointer as the locator, but they don't have to.  I think they may use a different form when it's an external reference.  Your algo may still work here, but it's a good idea to check on the JSON Ref specification.

com.reprezen.swagedit/src/com/reprezen/swagedit/validation/SwaggerError.java
```diff
@@ -109,20 +128,66 @@ public String getMessage() {
...
+		protected String getHumanFriendlyText(String location) {
+			JsonNode swaggerSchemaNode = findNode(location);
+			if (swaggerSchemaNode == null) {
+				return location;
+			}
+			JsonNode title = swaggerSchemaNode.get("title");
+			if (title != null) {
+				return title.asText();
+			}
+			JsonNode description = swaggerSchemaNode.get("title");
```
tedepstein: Was this supposed to be `.get("description")` ?  I'm not sure using the description is a good idea, because they can be too long for easy display in a format like this.  We might be better off putting titles in wherever they might be omitted from the original swagger schema.

tedepstein: I checked the Swagger Schema, and verified that all uses of the `oneOf` constraint use named schemas from `/definitions`.  So it should be safe to rely on this.

There are two uses of anyOf in "schema":

```
        "additionalProperties": {
          "anyOf": [
            {
              "$ref": "#/definitions/schema"
            },
            {
              "type": "boolean"
            }
          ],
```

and:

```
        "items": {
          "anyOf": [
            {
              "$ref": "#/definitions/schema"
            },
            {
              "type": "array",
              "minItems": 1,
              "items": {
                "$ref": "#/definitions/schema"
              }
            }
          ],
```

Also one more in "format":

```
    "format": {
      "anyOf": [
        {
          "title": "user-defined format",
          "type": "string"
        },
        {
          "title": "pre-defined format",
          "type": "string",
            "enum": [
              "int32",
              "int64",
              "float",
              "double",
              "byte",
              "binary",
              "date",
              "date-time",
              "password"                    
            ]
        }
      ]   
    }
```

I'm checking in a suggested change that adds titles to these cases.

tedepstein:
Passed code review.

tfesenko:
@tedepstein , a screenshot with the `title` you've just provided:
screen shot 2016-03-04 at 6 26 26 pm

tedepstein:
@tfesenko , I would suggest adding a couple of unit tests to prevent future failures:
1. Verify that all `$ref:` values in the schema yield a non-empty string when processed with your name extractor function (substring after the last `/` slash character).
2. Verify that all `allOf` or `anyOf` candidate subschemas are $refs (already verified in test 1) or have a "title" property.

tedepstein:
Test results:
- Logged an issue re: unsigned code warning
- Invalid parameter test passed: image
- additionalProperties test passed: 
  image
- array/items test passed:
  image
- format test passed, sort of.  It didn't list the two alternatives (user-defined format string vs. pre-defined enum format string).  Maybe because the validator combines the `type` validation on these two cases, and fails generally on an invalid type.  In any case, I am happy enough with this error message:
  image

tedepstein:
Passed QA.  Merging.

Issue #104, opened 2016-02-26
Improvements to Templates


tedepstein:
Collecting into a single issue for convenience.  We can break this into multiple issues if needed:
- **Wrong indent** - When I used the DELETE method template, it put extra space on the property lines
  ![image](https://cloud.githubusercontent.com/assets/692813/13368065/509388a6-dcb5-11e5-800e-1864b6e25591.png)
  
  The POST method had only one space on the property lines:
  
  ![image](https://cloud.githubusercontent.com/assets/692813/13368083/777d86ce-dcb5-11e5-8a93-553a97e0441a.png)
- **Templates should be complete** The templates included are not complete or realistic.  This is OK as a means of proving that template selection works, but a "placeholder" template like this could give users a bad impression.

tfesenko:
@tedepstein , you mentioned that you had links to good resources about best practices. Can you please share them with me?

tedepstein:
@tfesenko , please see comments on these issues:

https://modelsolv.atlassian.net/browse/ZEN-1051
https://modelsolv.atlassian.net/browse/ZEN-2360

tedepstein:
Also, @tfesenko , I have some links that I give to people  who ask about introductory material for REST.  I think these two might be helpful:
http://code.tutsplus.com/tutorials/a-beginners-guide-to-http-and-rest--net-16340
http://www.narwhl.com/

I would stay away from hypermedia and HATEOAS for now, because they're not supported in Swagger–OpenAPI.

PR #114, opened 2016-03-09
#104 - Improvements to Templates

tfesenko:


tedepstein:
@tfesenko , working on this, but will have to come back to  it later.

tfesenko:
Merging the PR as it definitely provides improvements, but leaving the issue open so we could return to it later.

Issue #117, opened 2016-04-20
Warn on duplicate properties


tedepstein:
Between, JSON, YAML and Swagger, there are some different rules and guidelines about duplicate property names; and implementations vary.

But a duplicate property name is almost always a user error, so SwagEdit should warn in this instance.

tedepstein:
@tfesenko, also flagging this as high priority.

ghillairet:
The branch task/117 contains a preliminary implementation to support warning on duplicate keys. It works on objects. Need to see what other kind of values should be supported (ex duplicate items in arrays?).

![screen shot 2016-05-19 at 16 06 02](https://cloud.githubusercontent.com/assets/148045/15396420/5776ac6c-1ddc-11e6-9dad-239eb1dcfd9a.png)

tfesenko:
@tedepstein , we need your feedback on phrasing, see the screenshot above

tedepstein:
I'd recommend the following error message:

> Duplicate key: the object [object name] has more than one property named [property name]

tedepstein:
@ghillairet asked:

> Need to see what other kind of values should be supported (ex duplicate items in arrays?).

I think this should be limited to duplicate property names in an object.  Duplicate array values are allowed, and should not cause a problem.

tfesenko:
The PR merged, but I am leaving this issue open.
The suggested phrasing for the error message was :

> Duplicate key: the object [object name] has more than one property named [property name]

The actual is: 

> error_duplicate_keys = object has duplicate keys: %s
> <img width="346" alt="screen shot 2016-05-30 at 8 37 04 pm" src="https://cloud.githubusercontent.com/assets/644582/15660168/8001dc2e-26a6-11e6-9fa7-d0fe65863803.png">

ghillairet:
I changed the message to `Duplicate key: the object has more than one property named %s`
It does not include the object name, how should we named an object?

tedepstein:
I think we can leave this as you've done it, without an object name.

tfesenko:
<img width="501" alt="screen shot 2016-06-02 at 6 29 47 pm" src="https://cloud.githubusercontent.com/assets/644582/15763173/03885898-28f0-11e6-8f6f-a85a436ee66f.png">

PR #124, opened 2016-05-19
[#117] Warn on duplicate properties

ghillairet:
See #117

### DIFF
--- a/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/ValidatorTest.xtend
+++ b/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/ValidatorTest.xtend
@@ -217,6 +217,31 @@ class ValidatorTest {
 	}
 
 	@Test
+	def void shouldNotWarnOnDuplicateKeys_InDifferentObjects() {
+		val content = '''
+			swagger: '2.0'
+			info:
+			  version: 0.0.0
+			  title: Simple API
+			paths:
+			  /foo/{bar}:
+			    get:
+			      responses:
+			        '200':
+			          description: OK
+			    put:
+			      responses:
+			        '200':
+			          description: OK
+		'''
+
+		document.set(content)
+		val errors = validator.validate(document)
+
+		assertEquals(0, errors.size())
+	}
+
+	@Test
 	def void shouldWarnOnDuplicateKeys_InsideObjects() {
 		val content = '''
 			swagger: '2.0'
@@ -272,6 +297,33 @@ class ValidatorTest {
 			new SwaggerError(11, IMarker.SEVERITY_WARNING, String.format(Messages.error_duplicate_keys, "in"))
 		))
 	}
-}
 
+	@Test
+	def void shouldWarnOnDuplicateKeys_InsidePaths() {
+		val content = '''
+			swagger: '2.0'
+			info:
+			  version: 0.0.0
+			  title: Simple API
+			paths:
+			  /foo/{bar}:
+			    get:
+			      responses:
+			        '200':
+			          description: OK
+			      responses:
+			        '201':
+			          description: OK
+		'''
+
+		document.set(content)
+		val errors = validator.validate(document)
+
+		assertEquals(2, errors.size())
+		assertThat(errors, hasItems(
+			new SwaggerError(8, IMarker.SEVERITY_WARNING, String.format(Messages.error_duplicate_keys, "responses")),
+			new SwaggerError(11, IMarker.SEVERITY_WARNING, String.format(Messages.error_duplicate_keys, "responses"))
+		))
+	}
+}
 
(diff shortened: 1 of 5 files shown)


com.reprezen.swagedit/src/com/reprezen/swagedit/validation/Validator.java
```diff
@@ -56,21 +66,85 @@
...
+			for (NodeTuple tuple: n.getValue()) {
+				Node keyNode = tuple.getKeyNode();
+
+				if (keyNode.getNodeId() == NodeId.scalar) {
+					String key = ((ScalarNode) keyNode).getValue();
+					if (duplicates.containsKey(key)) {
+						duplicates.get(key).add(keyNode);
+					} else {
+						duplicates.put(key, Sets.newHashSet(keyNode));
+					}
```
tfesenko: I don't know why you don't like Guava, but sometimes it's very useful. A `Multimap` would save us a four lines of the if-else here:

``` java
Multimap<String, Node> duplicates = HashMultimap.<String, Node>create();
<...>
duplicates.put(key, keyNode);
```

com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/ValidatorTest.xtend
```diff
@@ -187,4 +189,89 @@ class ValidatorTest {
...
+
+		document.set(content)
+		val errors = validator.validate(document)
+
+		assertEquals(2, errors.size())
+		assertThat(errors, hasItems(
+			new SwaggerError(10, IMarker.SEVERITY_WARNING, String.format(Messages.error_duplicate_keys, "in")),
+			new SwaggerError(11, IMarker.SEVERITY_WARNING, String.format(Messages.error_duplicate_keys, "in"))
+		))
+	}
```
tfesenko, line 299: I would also add a test for a case when the value is an object, for example here `responses` defined twice within the same pathItem:

``` yaml
swagger: '2.0'
info:
  version: 0.0.0
  title: Simple API
paths:
  /foo/{bar}:
    get:
      responses:
        '200':
          description: OK
      responses:
        '201':
          description: OK
```

It works as expected - validator shows a warning, but we don't have a test for it.

com.reprezen.swagedit/src/com/reprezen/swagedit/validation/Validator.java
```diff
@@ -56,21 +66,85 @@
...
+			for (String key: duplicates.keySet()) {
+				Set<Node> values = duplicates.get(key);
+				if (values.size() > 1) {
+					for (Node duplicate: values) {
+						errors.add(new SwaggerError(
+								duplicate.getStartMark().getLine() + 1, 
+								IMarker.SEVERITY_WARNING, 
+								String.format(Messages.error_duplicate_keys, key)));
+					}
+				}
```
tfesenko, line 138: I am thinking of a burden on performance. We have two passes here:
1. to create a map
2. to check that every key has a single value.
@ghillairet , let's think how we can improve it. Maybe, use a map that would throw an exception if you add a value for the same key twice and catch an exception.

com.reprezen.swagedit/src/com/reprezen/swagedit/validation/Validator.java
```diff
@@ -56,21 +66,85 @@
...
+		Set<SwaggerError> errors = Sets.newHashSet();
+		
+		if (node.getNodeId() == NodeId.mapping) {
+			MappingNode n = (MappingNode) node;
+
+			Map<String, Set<Node>> duplicates = Maps.newHashMap();
+			for (NodeTuple tuple: n.getValue()) {
+				Node keyNode = tuple.getKeyNode();
+
+				if (keyNode.getNodeId() == NodeId.scalar) {
```
tfesenko: I wonder why we need a check for `scalar` here

com.reprezen.swagedit.tests/src/com/reprezen/swagedit/tests/ValidatorTest.xtend
tfesenko, line 299: Thanks for the tests!

com.reprezen.swagedit/src/com/reprezen/swagedit/validation/Validator.java
ghillairet, line 138: If you throw the exception on the first key encounter twice, how do you handle cases where an object has multiple different duplicate keys?
What I can do is collect all duplicates in a single map and resolve duplicates once after the document is parsed.

com.reprezen.swagedit/src/com/reprezen/swagedit/validation/Validator.java
```diff
@@ -101,50 +97,65 @@
...
-								String.format(Messages.error_duplicate_keys, key)));
-					}
+	/*
+	 * This method iterates through the YAML tree to collect the pairs of Node x String 
+	 * representing an object and one of it's keys. Each pair is associated to a Set of 
+	 * Nodes which contains all nodes being a key to the pair's Node and having for value 
+	 * the pair's key.
+	 * Once the iteration is done, the resulting map should be traversed. Each pair having 
+	 * more than one element in its associated Set are duplicate keys. 
+	 */
```
tfesenko, line 129: This is a comment, not a Java doc. I wonder if it was your intention.
Similar with other methods in this class.

tfesenko:
Code looks good!
the performance is good too - it took 1-2 milliseconds on a valid 2800 line model to pass `checkDuplicateKey`

Issue #143, opened 2016-06-21
Should validate on open or project initialization, should not require edit


tedepstein:
Some validations don't seem to happen until the user makes an edit to a file.  This was discussed specifically in the context of reference validation in SwagEdit, but may be an issue with other kinds of validation as well.  Ideally, validation should happen on project initialization, so the user sees validation markers in Project Explorer or Navigator views, without having to open files.

If this is not feasible, it should happen on opening the file in the editor.  Currently, validation doesn't seem to happen until the first edit, and this is not intuitive to users.

tedepstein:
Further notes from @andylowry:

> The editor shows an error marker only after the file has been edited, which is something Tanya mentioned a few days ago in another context and seems like a bug to me. I would expect the file content to be run through validation immediately upon opening. (E.g. the Java editor does this: create bad java, save it, close it, delete the markers attached to the file, and reopen; you'll see markers reinstated. Same process with a swagger file results in no markers in the reopened editor.)

tedepstein:
Important to fix this as part of the fix for [ZEN-2407](https://modelsolv.atlassian.net/browse/ZEN-2407)

PR #145, opened 2016-07-06
[#143] Should validate on open or project initialization, should not …

ghillairet:
…require edit

### DIFF
--- a/com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerEditor.java
+++ b/com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerEditor.java
@@ -84,7 +84,7 @@ public class SwaggerEditor extends YEdit {
 					@Override
 					public void run() {
 						document.onChange();
-						validate();
+						validate(false);
 					}
 				});
 			}
@@ -184,7 +184,7 @@ public class SwaggerEditor extends YEdit {
 			if (document != null) {
 				document.addDocumentListener(changeListener);
 				// validate content before editor opens
-				validate();
+				validate(true);
 			}
 		}
 	}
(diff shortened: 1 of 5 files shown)


com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerEditor.java
```diff
@@ -181,6 +183,8 @@ protected void doSetInput(IEditorInput input) throws CoreException {
 			IDocument document = getDocumentProvider().getDocument(getEditorInput());
 			if (document != null) {
 				document.addDocumentListener(changeListener);
+				// validate content before editor opens
+				validate();
```
tfesenko: Nice

tfesenko:
Malformed YAML is not detected, probably because `yamlError` is not initialized on `com.reprezen.swagedit.editor.SwaggerEditor.doSetInput(IEditorInput)` yet as it's initialized on `com.reprezen.swagedit.editor.SwaggerDocument.onChange()`

To create a malformed YAML, I took a regular Swagger model and shifted one of the fields.

ghillairet:
The last commit fixes that problem, please have a look.

com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerEditor.java
```diff
@@ -282,16 +282,16 @@ public void updateFoldingStructure(List<Position> positions) {
...
 
 	@Override
 	public void doSaveAs() {
 		super.doSaveAs();
-		validate();
+		validate(false);
 	}
 
-	protected void validate() {
+	protected void validate(boolean onOpen) {
```
tfesenko, line 298: `validate(true)` is used only once - on `doSetInput()`, in all other cases it's `validate(false)`. You could have left the default implementation like this:

``` java
protected void validate() {
  validate(false);
}
protected void validate(boolean onOpen) { <...>}
```

ghillairet, line 298: True, I updated the PR.

tfesenko:
Looks good!

Issue #142, opened 2016-06-21
Failure to detect broken references


tedepstein:
Saul posted a multi-file project with some broken references that were not detected by validation.  These were mostly due to case-sensitivity, but reference validation should be case-sensitive, otherwise Swagger specs will not be portable.

ghillairet:
I have made a PR for this, see https://github.com/RepreZen/SwagEdit/pull/146
@tfesenko Please have a look to see if it does what is needed.

PR #146, opened 2016-07-06
[#142] Failure to detect broken references

ghillairet:
This commit modifies the JsonReference validation process by checking that the file the JsonReference URI is pointing at exists and is accessible. This is necessary to handle case sensitivity. The validation previously worked because validation was done on URIs and did not check that the file could be open. Hyperlinks to files with a different case than the one in the JsonReference previously failed but validation passed. Now, the JsonReference URI and the file name should have same case.

### DIFF
--- a/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/mocks/Mocks.java
+++ b/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/mocks/Mocks.java
@@ -1,0 +1,67 @@
+package com.reprezen.swagedit.mocks;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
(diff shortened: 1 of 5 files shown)


com.reprezen.swagedit.tests/src/com/reprezen/swagedit/mocks/Mocks.java
```diff
@@ -0,0 +1,67 @@
...
+					}
+					@Override
+					public JsonNode getDocument(URI uri) {
+						return entries.get(uri);
+					}
+				});
+				return ref;
+			};
+		};
+	}
```
tfesenko, line 66: Thanks

tfesenko:
Passed code review.
Works as expected on OS X

Issue #150, opened 2016-07-13
External reference shows warning


tedepstein:
Related to [ZEN-2780](https://modelsolv.atlassian.net/browse/ZEN-2780).

In the attached workspace, I used the new Code Assist cycling feature in SwagEdit to insert a reference to a schema object defined in an external file. Uber.yaml has a reference to the Pets schema object in Pet Store.yaml. The reference looks correct to me, but SwagEdit shows a warning, and the Doc View shows a rendering error.

[RepreZen workspace cycling.zip](https://github.com/RepreZen/SwagEdit/files/361967/RepreZen.workspace.cycling.zip)

<img width="967" alt="reprezen - ubermodelsuber yaml - reprezen api studio_2016-07-13_12-31-38" src="https://cloud.githubusercontent.com/assets/692813/16811595/20dfd8cc-48f7-11e6-9381-647c674cd36c.png">

tfesenko:
As @andylowry pointed in ZEN-2780:

> Reference is invalid. The "URL" part of the reference must conform to URI syntax, which does not permit space characters. The relative URL should be url-escpaped when creating the URL (i.e. this is something content-assist needs to pay attention to). Change "Pet Store" to "Pet%20Store" and all works as expected.

Code-assist should insert a valid URL-escaped strings for refs

tedepstein:
So, as discussed, there are two things to be fixed in SwagEdit:
1. Code Assist should replace spaces with %20.
2. References with illegal URI syntax should be treated as an error, not a warning. This is different from legal URI that cannot be resolved, or that does not resolve to a valid object of the expected type. Unresolvable references are still warnings; illegal URIs are errors.

That latter issue is important so that the live views don't try to render an invalid swagger doc, throw and exception, and display the last-resort "Rending Error."

ghillairet:
PR https://github.com/RepreZen/SwagEdit/pull/151

tfesenko:
The major part is done, please correct the error message to close this issue.

ghillairet:
The message has been updated in this PR https://github.com/RepreZen/SwagEdit/pull/153

PR #153, opened 2016-07-15
[#150] External reference shows warning

ghillairet:
Fix error message and code cleaning.

### DIFF
--- a/com.reprezen.swagedit/src/com/reprezen/swagedit/assist/SwaggerContentAssistProcessor.java
+++ b/com.reprezen.swagedit/src/com/reprezen/swagedit/assist/SwaggerContentAssistProcessor.java
@@ -100,21 +100,7 @@ public class SwaggerContentAssistProcessor extends TemplateCompletionProcessor i
 		String bindingKey = bindingService
 				.getBestActiveBindingFormattedFor(IWorkbenchCommandConstants.EDIT_CONTENT_ASSIST);
 
-		String context;
-		switch(ContextType.get(currentPath)) {
-			case PATH_ITEM:
-				context = "path items";
-				break;
-			case PATH_PARAMETER:
-				context = "parameters";
-				break;
-			case PATH_RESPONSE:
-				context = "responses";
-				break;
-			default:
-				context = "schemas";
-				break;
-		}
+		String context = ContextType.get(currentPath).label();
 
 		return new String[] {
 				String.format(Messages.content_assist_proposal_project, bindingKey, context),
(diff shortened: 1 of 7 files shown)


com.reprezen.swagedit/src/com/reprezen/swagedit/assist/SwaggerContentAssistProcessor.java
```diff
@@ -100,21 +100,7 @@ public SwaggerContentAssistProcessor(ContentAssistant ca) {
...
-				context = "parameters";
-				break;
-			case PATH_RESPONSE:
-				context = "responses";
-				break;
-			default:
-				context = "schemas";
-				break;
-		}
+		String context = ContextType.get(currentPath).label();
```
tfesenko, line 103: Thanks

com.reprezen.swagedit/src/com/reprezen/swagedit/messages.properties
```diff
@@ -21,9 +21,9 @@ error_additional_properties_not_allowed = object has properties %s which are not
 error_required_properties = object has missing required properties %s
 error_duplicate_keys = Duplicate key: the object has more than one property named %s
 error_cannot_read_content = Unable to read content. It may be invalid YAML
-error_invalid_reference = Invalid Reference - SwagEdit was unable to resolve the reference. \n\
+error_missing_reference=Invalid Reference - SwagEdit was unable to resolve the reference. \n\
 The value must be a valid JSON Reference (for external references) or JSON Pointer (for local references), and must resolve to an object of the expected type.
-
+error_invalid_reference=Invalid Reference Syntax - The referenced path or URI may contain invalid characters.
```
tfesenko: In https://github.com/RepreZen/SwagEdit/pull/151#discussion_r70855192 @tedepstein wrote: 

> It would be better to change the first line to this: <...>
> The second line can remain the same.

I interpret it as that we need a two-line message:

> Invalid Reference Syntax - The referenced path or URI may contain invalid characters.  
> The value must be a valid JSON Reference (for external references) or JSON Pointer (for local references), and must resolve to an object of the expected type.

@tedepstein , can you please confirm that this is the message you suggested?

While the new code has only the first line and omits the second line:
screen shot 2016-07-15 at 9 49 09 am

ghillairet: I added the missing part of the message.

tfesenko:
Looks good, @ghillairet 
screen shot 2016-07-15 at 10 50 19 am

tedepstein:
Looks good!

Issue #125, opened 2016-05-28
Invalid JSON Example shown at top level as "object has missing required properties"


tedepstein:
Reported by Saul in [support ticket #111](https://support.reprezen.com/helpdesk/tickets/111):

In the project attached to the support ticket, the file `/models/PathItems/ropeSwagger_company_&companyId_serviceable_&serviceableId_products.json.yaml` has invalid json in its example, starting on line 46.

It's shown as a schema validation problem:

```
Multiple markers at this line
- object has missing required properties "info", "paths", "swagger"
- 44 changed lines
```

The invalid JSON seems to have made SwagEdit unable to parse the file correctly, so the error message doesn't point to the issue correctly.  

It would be better if we could report more accurately what caused the problem.

tedepstein:
Emerged again in https://support.reprezen.com/helpdesk/tickets/124

tedepstein:
Discussed in [ZEN-2407](https://modelsolv.atlassian.net/browse/ZEN-2407).  Marking as high priority.  A parse failure needs to be identified as such, otherwise it becomes a support issue.

ghillairet:
What was the problem in the JSON example? I made a screencast showing the errors when invalid JSON is present.

![json](https://cloud.githubusercontent.com/assets/148045/16876592/bef9634e-4aa4-11e6-8500-ae1edbe610cf.gif)

tfesenko:
I am able to reproduce this issue:
<img width="428" alt="screen shot 2016-07-15 at 10 22 59 am" src="https://cloud.githubusercontent.com/assets/644582/16877460/2e79df66-4a76-11e6-92d4-49e3cb5a4b5d.png">

@ghillairet I will try to create a simplified model for you to demonstrate this issue.

tfesenko:
This model creates yields an "Unable to read contents error". It contains a valid JSON wrapped in extra curly braces.
Note that the error message is more helpful that the one which was originally reported.

The error is shown on the first line:
<img width="714" alt="screen shot 2016-07-15 at 10 42 56 am" src="https://cloud.githubusercontent.com/assets/644582/16878056/0517d26a-4a79-11e6-85aa-9655c839e3d5.png">

``` yaml
---
swagger: "2.0"
info:
  description: Tax Blaster
  version: 1.0.0
  title: TaxBlaster
host: taxblaster.com
basePath: /api
schemes:
- http

tags:
- name: TaxFilingObject
[...]

PR #157, opened 2016-07-15
[#125] Invalid JSON Example shown at top level as "object has missing…

ghillairet:
… required properties"

Make sure JSON parse exceptions are rendered as markers.

### DIFF
--- a/com.reprezen.swagedit/src/com/reprezen/swagedit/validation/SwaggerError.java
+++ b/com.reprezen.swagedit/src/com/reprezen/swagedit/validation/SwaggerError.java
@@ -22,6 +22,7 @@ import org.eclipse.core.resources.IMarker;
 import org.yaml.snakeyaml.error.MarkedYAMLException;
 import org.yaml.snakeyaml.error.YAMLException;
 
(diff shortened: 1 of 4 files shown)


com.reprezen.swagedit/src/com/reprezen/swagedit/validation/SwaggerError.java
```diff
@@ -61,6 +62,17 @@ public SwaggerError(YAMLException exception) {
...
+		this.level = IMarker.SEVERITY_ERROR;
+		this.message = exception.getMessage();
+
+		if (exception.getLocation() != null) {
+			this.line = exception.getLocation().getLineNr();
+		} else {
+			this.line = 1;
+		}
+	}
+
```
tfesenko, line 75: Thanks!

tfesenko:
The line is shown correctly, which is already a great help, but the message looks a too technical and at the same time does not explain the problem:

for my model:
screen shot 2016-07-15 at 12 07 46 pm

and for the original model:
screen shot 2016-07-15 at 12 08 09 pm

Issue #158, opened 2016-07-15
Swagger spec which uses special characters shows an error dialog


tfesenko:
We have a Windows user who uses special characters (from Icelandic alphabet). 
The user reported the following error (see https://support.reprezen.com/helpdesk/tickets/133):

> ‘Update SwagEdit validation markers‘ has encountered a problem.
> An internal error occurred during: “Update SwagEdit validation markers“
> 
> When I press details I get some extra info that seems to point to some encoding issue.
> I can‘t seem to find anything on google regarding this issue so I really need your help.
> I will send a screenshot as well as the .yaml  code as it is right now.
> 
> I am running on windows.
> 
> I had been writing the project for a couple hours without any problems when this suddenly happened.
> I am not entirely sure what I did but it felt like this happened when I hit CTRL + S to save the project.

I was able to reproduce it on Mac. Note that original encoding is "iso-8859-1":

```
file -I creditcards.yaml
creditcards.yaml: text/plain; charset=iso-8859-1
```

Also, the spec is invalid because the `host` is not specified:

``` yaml
info:
  description: yoyoyo
  version: 1.0.0
  title: CreditCards
host: 
```

This has high priority

tedepstein:
Heard back from Egill:

> Thanks for quick responses, great service.  It worked for me to set the default to UTF-8 and writing Icelandic letters even seems to work after that.

I'm still not sure what this means for us though:  should we change the default encoding for Windows installation and possibly other platforms, and/or just recommend this as a workaround if users encounter the problem?  Or is there a better solution?

Also, I assume we need to do something with exception handling so that users don't see repeated modal error messages whenever they type in the editor.

I've attached the problem source file here.

[creditcards.zip](https://github.com/RepreZen/SwagEdit/files/369016/creditcards.zip)

ghillairet:
The problem looks like JSON only allows UTF encodings, the special character contain in the sample projects were encoded in iso-8859-1 which causes a parsing exception to be thrown.

I made the changes to force the parsing of swagger document using UTF-8 encoding.

I don't know what's the default file encoding in Eclipse for Windows, but we should encourage users to set it to UTF-8. This can be done in Eclipse > Preferences > General > Workspace > Text file encoding. 

Link to PR https://github.com/RepreZen/SwagEdit/pull/160

tfesenko:
I suggest to limit the scope of this issue to the following:
[...]

PR #160, opened 2016-07-21
[#158] Swagger spec which uses special characters shows an error dialog

ghillairet:


com.reprezen.swagedit/src/com/reprezen/swagedit/json/JsonDocumentManager.java
```diff
@@ -90,6 +94,34 @@ public JsonNode getDocument(URL url) {
...
+     * object mapper passed as parameter.
+     * 
+     * It also forces the content of the URL to be read with UTF-8 encoding, considering that only UTF encodings are
+     * valid encodings for JSON inputs.
+     */
+    protected JsonNode doRead(ObjectMapper mapper, URL url) throws Exception {
+        try (InputStream stream = url.openStream()) {
+            try (BufferedReader reader = new BufferedReader(new InputStreamReader(stream, "UTF-8"))) {
+                return mapper.readTree(reader);
+            } catch (CharConversionException e) {
```
tfesenko, line 108: Thanks!

com.reprezen.swagedit/src/com/reprezen/swagedit/json/JsonDocumentManager.java
```diff
@@ -90,6 +94,34 @@ public JsonNode getDocument(URL url) {
...
+                throw new Exception(e);
+            }
+        } catch (IOException e) {
+            throw new Exception(e);
```
tfesenko, line 112: I wonder why we need to wrap existing `IOException` and `CharConversionException` into a generic `Exception`...

com.reprezen.swagedit/src/com/reprezen/swagedit/json/JsonDocumentManager.java
```diff
@@ -61,24 +65,24 @@ public JsonNode getDocument(URL url) {
         JsonNode document;
         if (url.getFile().endsWith("json")) {
             try {
-                document = mapper.readTree(url);
+                document = doRead(mapper, url);
             } catch (Exception e) {
                 document = null;
```
tfesenko, line 70: Current behaviour on this task branch: 
When I open a model with non-UTF-8 symbols the symbols are silently replaced by the `�`, nothing indicates it - no entries in the error log, no markers in the problem view. In many European languages non-UTF symbols are frequent and loosing them is not what users expect.

While not showing an error dialog is an definite improvement, probably we still should report that something went wrong. 
@tedepstein we need your feedback here on how to indicate that a file has unexpected non-UTF-8 symbols and communicate why they are not allowed. I will list the most obvious options:
1. Display an error dialog on file opening. Yes, this is exactly what we removed in this pull request, but we can make the error dialog more informative and show it only once - on file opening.
2. Show warnings markers. Ideally, they should be shown on the lines where we meet invalid symbols, but I can't tell you effort estimate for this.
3. Report it in the error log. It's probably the easiest to implement, but it's also the least visible.

tedepstein:
@tfesenko and @ghillairet , I have to confess that I don't fully understand how character encoding schemes work, and I would need to do at least a little background reading before I could properly suggest how we should go forward.    Unfortunately I don't have time to do that right now.

@andylowry , I'd appreciate your input here.

andylowry:
@tedepstein, @tfesenko 

First a bit of background: UTF-8 is an "encoding scheme" that is used to encode the Unicode character set. The Unicode character set is intended to cover every single character that has ever been used in any human language, and UTF-8 is capable of encoding every one of them. So it's not really accurate to say that in European languages non-UTF symbols are frequent. It would be more meaningful to say that non-Unicode characters are frequent, but that would also be false.

The problem is that Unicode is not the only character set that has been devised in the world, and UTF-8 (and UTF-16 and UTF-32, all capable of encoding any Unicode text) are not the only encoding schemes. Increasingly, systems are moving to Unicode and its encoding schemes as standard practice, but there are still lots of systems and software that either default to other choices or can be configured to use other choices.

One of the features of Unicode is that its first 128 code positions are exactly aligned with US-ASCII, and the UTF encoding of these characters is also precisely aligned with US-ASCII. So, e.g. a byte with the decimal value 65 is represents the letter "A" in US-ASCII, and likewise in UTF-8. In fact, most other encoding schemes that have been developed - including those catering to European language needs (accented characters, etc.) as well as non-Roman languages (Greek, Arabic, Chinese, etc.) share this strict conformity with US-ASCII in their first 128 character positions (0-127). This has made the underlying tower-of-babel far less of a practical problem than it otherwise would have been, but it has left ample room for pain.

When any piece of software attempts to understand a piece of text, it's looking, of course, at a sequence of bits, and it needs to adopt an encoding scheme to attempt to turn the bits into a sequence of characters in some character set. And that's where the difference between an encoding and a character set shows up. A character set is a 1:1 mapping of integers to glpyhs. An encoding is a set of rules for translating between a bit sequence and a the characters in a given character set. Encoding schemes are generally associated with character sets, but there may be multiple encoding schemes for the same character set. That's the case, for example, with Unicode.

Unicode has four encoding schemes that have not become completely obsolete:
- UTF-8 is a variable-width encoding scheme and is the most common scheme for external storage of Unicode text. It uses anywhere from one to four bytes to encode each character in the sequence, with the first byte of the sequence indicating - via its high order bits - how long the sequence is. E.g. if the highest bit in the first byte is zero, then this is a one-byte encoding, and the other 7 bits identify a character in the range 0-127. This is the part of Unicode that aligns with US-ASCII. If the first two bits in the first byte are `11` then this is a two-byte sequence, and so on. In multi-byte sequences, each subsequent byte must have `10' as its high-level bits, so it is possible (and in fact quite likely) for data that is _not_ intended to be UTF-8 encoded text to present encoding failures - points in the bit stream where the data does not correspond to any possible UTF-8 encoding.
- UCS-2 is an encoding scheme in which each character is encoded in two bytes. I believe it was the encoding initially used by Java internally, but Java now uses UTF-16. UCS-2 is not capable of representing the entire Unicode character set.
- UTF-16 uses either one or two 16-bit values to represent a Unicode character, and I believe that every valid UCS-2 character encoding is identical to its UTF-16 encoding, but I'm not completely sure of that.
- UTF-32 uses a single 32-bit value to encode every Unicode character. The gcc compiler - and I believe recent Python implementations - use UTF-32 internally.

So how's a piece of software to know what it's looking at? In the case of XML, there's a header that's carefully constrained to contain only US-ASCII characters, which - as mentioned above - aligns with nearly every character scheme in use. That header includes an "encoding" attribute that identifies the encoding. However it's very common for programmers to screw this up and just mark every XML file they produce as "UTF-8" without taking care to actually use that encoding for the content.

Some programs - like Eclipse - can be configured to assume one of various character encodings. Some, like emacs, have rich features for changing the encoding of a file being edited, translating the data accordingly.

Some programs will attempt to discern the encoding by examining the data, but this is difficult and never fool-proof.

The JSON spec (RFC 4627) includes the following:

>  JSON text SHALL be encoded in Unicode.  The default encoding is  UTF-8.
> 
>  Since the first two characters of a JSON text will always be ASCII  characters [RFC0020], it is possible to determine whether an octet stream is UTF-8, UTF-16 (BE or LE), or UTF-32 (BE or LE) by looking at the pattern of nulls in the first four octets.

The YAML specification also specifies Unicode character set, and it requires processors to support both UTF-8 and UTF-16 encodings. It mentions that processors that wish to be JSON-compliant must also support UTF-32.

This means that a YAML editor like YEdit (and hence SwagEdit) is required to understand UTF-8, UTF-16, and UTF-32 encodings. They are not required to understand any other encoding, and probably do not. When faced with other encodings (or non-textual data), all they can really do is _attempt_ to interpret the data as a sequence of Unicode characters and fail as gracefully as possible when invalid bit sequences are encountered. I'm sure that's what's giving rise to the `�` characters Tanya reported.

So what can _we_ do to help our users? I would say it really comes down to a few practical things:
1. When invalid sequences are encountered, create a warning to that effect. This might be difficult to address in SwagEdit, and may require changes to YEdit. Here I'm talking about invalid encodings, not invalid YAML or Swagger.
2. Modify API Studio to ensure that the workspace default encoding is always UTF-8 in our RCS product.
3. Modify SwagEdit to check the workspace default encoding and offer some sort of warning dialog explaining the situation and offering to change the preference (should also have a "don't bother me about this again" checkbox).
4. Create a FAQ item explaining the issue and our recommended practice.

I don't think it's practical for us to try to do something smart with files that use other encodings. JSON & YAML files aren't marked with encodings like XML is (in fact, XML is fairly unusual in that respect - most file formats have no provision for declaring a character encoding). Asking the user would be weird and would probably be asking them a question they don't know how to answer. Attempting to automatically detect an encoding is far from fool-proof.

andylowry:
It's also worth understanding where the "Latin" encodings fit in this picture. There are several of these, all issued as variants of ISO 8859. These are all 8-bit encodings of character sets optimized for different European languages, and they became quite common long ago and have far from died off. Each one assigns several characters to range 128-255, i.e. bytes with their high order bit set.

Attempting to read a file in one of these encodings will fail when one of these high bytes is encountered, since by UTF-8 rules the byte will either be invalid or will introduce a multi-byte sequence that may be invalid even if its first byte is valid. Even if the byte and its successors happen to constitute a valid UTF-8 sequence, the result will certainly not be what's intended.

tedepstein:
Great explanation, Andy.  No doubt it took you some time, but you saved me a lot of headaches.

As for the recommendations:

> When invalid sequences are encountered, create a warning to that effect. This might be difficult to address in SwagEdit, and may require changes to YEdit. Here I'm talking about invalid encodings, not invalid YAML or Swagger.

Let's assess the difficulty of doing this.  If it's going to be really hard to detect these cases, we can live without the warning, I think.

> Modify API Studio to ensure that the workspace default encoding is always UTF-8 in our RCS product.

This default controls the encoding of files created in the workspace, and the default _decoding_ of files opened in editors, fed into parsers & compilers, etc.  Correct?  

This was the part I wasn't entirely sure about.  My concern is that it might cause problems in decoding files that exist in the user's environment.  I assume that the Eclipse developers gave some thought to the default encoding assumptions, and messing with this factory default could have unintended consequences.

Clearly, Eclipse allows you to explicitly set the encoding on a per-file basis, overriding the default.  It also allows the workspace default encoding to be specialized for certain content types:

image

It seems to me that it might be safer to leave the Eclipse workspace encoding as the default, which is platform-dependent and is determined by Eclipse, and just override the default for YAML and JSON.  For some reason I don't see JSON in the Content Types list, but I assume that's something we can figure out.

What do you think of that idea?  

> Modify SwagEdit to check the workspace default encoding and offer some sort of warning dialog explaining the situation and offering to change the preference (should also have a "don't bother me about this again" checkbox).

Sounds good, but we generally avoid using negatives in the wording of these checkboxes.  So the checkbox label should be more like "Verify encoding scheme when loading a file". Checked by default, and with an option to re-enable in SwagEdit preferences.

> Create a FAQ item explaining the issue and our recommended practice.

Good idea.  Starting with the very friendly explanation in your comments here.

andylowry:
@tedepstein, I was unaware of the per-content-type default encoding settings. I agree that's far preferable. I've just played around with it a bit, and it seems to handle things nicely. It turns out that eclipse keeps track of encoding settings for all resources and containers in a workspace. (Look at their properties to see this.) By default, containers (e.g. projects, folders) inherit their default encoding from their parent (and the workspace gets its default form preferences), and resources inherit their default from their immediate container unless their content type specifies a default encoding. You can manually change the encoding of any resource or container, but that doesn't recursively change contained resources and containers that currently show as inheriting their encoding. Instead, their encodings are retained and marked as non-inherited.

Or at least that's what I've gathered from a bit of experimentation. Seems pretty logical to me.

Here's a page that describes how plugins can define new content types: http://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2Fruntime_content_contributing.htm

It would seem advisable for us to create a content type for RAPD-ML files at the very least. The fact that - (at least  in my workbench) there is no JSON content type is surprising to me. We might want to define one, with UTF-8 as its default encoding, but I wonder if other available plugins also do so, and what would happen if more than one plugin tries to define the same content type? A quick search didn't yield any info on that last point.

We can also create "Swagger" subtypes of both YAML and JSON, and I think it certainly makes sense for SwagEdit to define these.

Whatever is responsible for creating the YAML content type in my workbench did not assign a default encoding, which it obviously should, given that the YAML spec requires Unicode character set.

It might also be a good idea to create an Xtend content type that sets utf-8 as a default encoding, since without that the guillemet characters in templates are not recognized.

Eclipse also supports the creation of "detectors" to decide whether a given stream of data corresponds to a particular content type. We might want to recast our current `LoadabilityTester` implementations as content type detectors.

tedepstein:
> It would seem advisable for us to create a content type for RAPD-ML files at the very least. 

Out of scope for SwagEdit, but I will open an issue in Jira for this.

> The fact that - (at least in my workbench) there is no JSON content type is surprising to me. We might want to define one, with UTF-8 as its default encoding...

Also, I think, separate from SwagEdit.  

> We can also create "Swagger" subtypes of both YAML and JSON, and I think it certainly makes sense for SwagEdit to define these.

Wondering if that's needed, or appropriate, because the UTF encoding requirements and recommendations are native to JSON and YAML.  So providing the YAML content type and setting its default encoding seems like something that belongs in YEdit.  

@tfesenko  and @ghillairet , would it be practical to open an issue in YEdit and address this as a PR to YEdit?  Of course we can maintain the fork ourselves, we don't need to wait for YEdit to merge the PR.  But it would be good to first get an indication from the YEdit author as to whether he recognizes the problem, and would be inclined to merge a fix.

We could consider a similar approach for JSON.  I will include this in a Jira issue.

> It might also be a good idea to create an Xtend content type that sets utf-8 as a default encoding, since without that the guillemet characters in templates are not recognized.

Also for consideration in API Studio, outside scope of SwagEdit.

> Eclipse also supports the creation of "detectors" to decide whether a given stream of data corresponds to a particular content type. We might want to recast our current LoadabilityTester implementations as content type detectors.

IIUC, ContentType determines default encoding (maybe among other things), and LoadabilityTester determines the icon decoration in file views and determines the default editor.  Correct?

In that case, I still think it makes sense to keep these separate.  Default content type of UTF-8 should apply to all YAML files, including, for example, the YAML files that API Studio uses for GenTarget configuration.  Swagger icon decoration and default editor assignment should only apply to Swagger YAML files.

andylowry:
@tedepstein The reason I thought Swagger subtypes for YAML and JSON content types would be worth considering in SwagEdit is the possibility of replacing our "loadability testers" with content type "detectors," which would make them more widely usable and efficient (due to eclipse' resource bookkeeping), and could simplify some of our code, e.g. in wizards. Agreed that the other two aspects of content type - file type associations and default encoding - would be fully covered in the YAML and JSON content types.

com.reprezen.swagedit/src/com/reprezen/swagedit/json/JsonDocumentManager.java
ghillairet, line 112: Yes I should change that.

tedepstein:
@andylowry , @tfesenko , this has been in limbo for a while.  I just had a quick read through this, and it seems like most of what was discussed after the last commit is possible follow-up work that could be extracted to separate issues, and prioritized independently.

@tfesenko , can you please review, extract issues as appropriate, and merge if you think it's OK in its current state?

Issue #166, opened 2016-07-28
NullPointerException if a reference has a space following the "#" (e.g. `# /definitions/Pet`)


tfesenko:
Open this model:

``` yaml
swagger: "2.0"
info:
  version: 1.0.0
  title: Swagger Petstore
host: petstore.swagger.io
basePath: /v1
schemes:
  - http
paths:
  /pet:
    get:
      summary: Info for a specific pet
      responses:
        200:
          description: success
          schema:
            $ref: "# /definitions/Pet"
definitions:
  Pet:
    properties:
      id:
        type: integer
        format: int64
```

A modal dialog will be shown reporting an NPE. The NPE is produced by the `resolved = doc.at(pointer);` in `com.reprezen.swagedit.json.references.JsonReference.resolve(URI)`.
No error marker added to the reference

The fix should incorporate the following:
1. Fix NPE
2. Report an error, not a warning, as in other cases with invalid ref strings
3. The modal dialog was caused by an exception inside an Eclipse `Job`. Make our Eclipse `Job`s used for validation more resilient to potential exception, at least, don't show a dialog.

Also see https://modelsolv.atlassian.net/browse/ZEN-2799

ghillairet:
I think this issue can be close, the error is correctly reported and there's no more any NPE.

PR #167, opened 2016-07-29
#166 NullPointerException if a reference has a space following the "#" (e.g. `# /definitions/Pet`)

tfesenko:
This PR addresses all three aspects of #166 
1. Fix NPE
2. Report an error, not a warning, as in other cases with invalid ref strings
3. The modal dialog was caused by an exception inside an Eclipse Job. Make our Eclipse Jobs used for validation more resilient to potential exception, at least, don't show a dialog.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ hs_err_pid*
 
 # maven build results
 target/
+
+# Xtend Gen folder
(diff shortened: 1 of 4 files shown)


ghillairet:
Looks good.

PR #176, opened 2016-08-10
Task/40 model nodes

ghillairet:
PR open only for discussions now.

Introducing a novel approach for parsing of document content.

The novel part has been done in the model package https://github.com/RepreZen/SwagEdit/tree/task/40-model-nodes/com.reprezen.swagedit/src/com/reprezen/swagedit/model. The parsing is now done by NodeDeserializer that uses Jackson streaming API to build a model representation of the document content. This model could be use by the different services offered by SwagEdit such as content assist, outline, etc...


tfesenko:
Note that this PR was targeting the `task/40`, not the `master`.

Issue #40, opened 2015-12-17
Implement Quick Outline


tedepstein:
It would be nice to have QuickOutline view implemented, so you press Ctrl+o (on Windows), type in a few characters, and find the definition you're looking for.

tfesenko:
Quick outline implementation in:
- JDT - http://git.eclipse.org/c/jdt/eclipse.jdt.ui.git/tree/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaOutlineInformationControl.java
- Xtext - https://github.com/eclipse/xtext/blob/master/plugins/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/outline/quickoutline/QuickOutlinePopup.java

One can think that you get a Quick Outline View for free if you already have a working Outline view, it's not true. Additional effort is needed, see links above.

ghillairet:
The problems I have noted while implementing the quick outline.
- Yedit provides an outline implementation that could have been reused, especially the class YAMLOutlineElement. Unfortunately this class has a constructor that takes a snakeyaml.Node as parameter and since Yedit and SwagEdit both ship with their own snakeyaml jars, trying to instantiate that class from SwagEdit causes a class loader exception.
- Lack of documentation for setting up a quick outline correctly in a text editor. I found what it looks like a good example in mylyn wikitext plugins. The interesting classes to look at are [MarkupSourceViewerConfiguration](https://github.com/eclipse/mylyn.docs/blob/master/org.eclipse.mylyn.wikitext.ui/src/org/eclipse/mylyn/wikitext/ui/editor/MarkupSourceViewerConfiguration.java), [MarkupSourceViewer](https://github.com/eclipse/mylyn.docs/blob/master/org.eclipse.mylyn.wikitext.ui/src/org/eclipse/mylyn/wikitext/ui/editor/MarkupSourceViewer.java) and [WikiTextSourceEditor](https://github.com/eclipse/mylyn.docs/blob/master/org.eclipse.mylyn.wikitext.ui/src/org/eclipse/mylyn/wikitext
  /ui/editor/WikiTextSourceEditor.java).

Since I could not reuse outline elements from YEdit, I had to implement my owns in SwagEdit. At this time the quick outline in SwagEdit looks like this.

[...]

PR #178, opened 2016-08-11
#40 Implement Quick Outline: better labels

ghillairet:
This PR extends the current quick outline with better labels(#40) and introduces a new outline that overrides the one provided by YEdit with content identical to the quick outline.
The (quick) outline uses the unified model to represent the content of swagger documents.


com.reprezen.swagedit.tests/src/com/reprezen/swagedit/editor/outline/AbstractNodeTest.xtend
```diff
@@ -0,0 +1,385 @@
...
+		val position = el.getPosition(doc)
+		// after foo:
+		assertEquals(4, position.offset)
+		assertEquals(0, position.length)
+
+		// should be first line
+		assertEquals(0, doc.getLineOfOffset(position.offset))
+	}
+	
+//	@Test
```
tfesenko, line 222: The `@Test` annotation is comment. Is it because the implementation of the **test** is incomplete or because of  the code? Any reason for not using `@Ignore`?

com.reprezen.swagedit.tests/src/com/reprezen/swagedit/editor/outline/AbstractNodeTest.xtend
```diff
@@ -0,0 +1,385 @@
...
+		val doc = new SwaggerDocument
+		doc.set(text)
+
+		val position = el.getPosition(doc)
+		// after foo:
+		assertEquals(4, position.offset)
+		assertEquals(0, position.length)
+
+		// should be first line
+		assertEquals(0, doc.getLineOfOffset(position.offset))
```
tfesenko, line 219: Thanks!

com.reprezen.swagedit.tests/src/com/reprezen/swagedit/editor/outline/OutlineStyledLabelProviderTest.xtend
```diff
@@ -44,18 +42,62 @@ class OutlineStyledLabelProviderTest {
...
+		val text = '''
+			objects:
+			  - name: hello
+			    value: world
+			  - name: hello
+			    value: world
+		'''
+		
+		val els = Model.parseYaml(text).root
+		assertEquals("objects", els.get(0).text)
```
tfesenko, line 84: Nice

com.reprezen.swagedit.tests/src/com/reprezen/swagedit/editor/outline/OutlineStyledLabelProviderTest.xtend
```diff
@@ -44,18 +42,62 @@ class OutlineStyledLabelProviderTest {
...
+		
+		// TODO
```
tfesenko, line 86: Let's create an issue for the TODO items, otherwise we can easily forget about them

com.reprezen.swagedit/src/com/reprezen/swagedit/editor/outline/OutlineStyledLabelProvider.java
```diff
@@ -56,35 +59,65 @@ public Styler getTextStyler() {
...
 
             cell.setText(styledString.toString());
             cell.setStyleRanges(styledString.getStyleRanges());
-            cell.setImage(getImage(getIcon((OutlineElement) element)));
+            cell.setImage(getImage(getIcon((AbstractNode) element)));
         }
     }
 
-    protected StyledString getSyledString(OutlineElement element) {
+    protected StyledString getSyledString(AbstractNode element) {
```
tfesenko, line 72: Typo: a missing `t` in `getSyledString`

com.reprezen.swagedit/src/com/reprezen/swagedit/editor/outline/QuickOutline.java
```diff
@@ -121,6 +122,7 @@ protected TreeViewer createTreeViewer(Composite parent) {
         treeViewer.addFilter(new NamePatternFilter());
         // Using ALL_LEVELS will cause editor to hang on large specs
         treeViewer.setAutoExpandLevel(2);
```
tfesenko, line 124: Good. Let's extract the `AutoExpandLevel` into a constant

com.reprezen.swagedit/src/com/reprezen/swagedit/model/AbstractNode.java
```diff
@@ -0,0 +1,150 @@
...
+            int startLine = location.getLineNr() - 1;
+            int offset = 0;
+            try {
+                offset = document.getLineOffset(startLine) + location.getColumnNr()
+                        + Strings.nullToEmpty(getProperty()).length();
+            } catch (BadLocationException e) {
+                return new Position(0);
+            }
+
+            return new Position(Math.max(0, offset), 0);
```
tfesenko: Because the length (the second parameter) is always null navigation did not work and Quick Outline was basically useless. I provided a simple fix in commit d90ea00390076f38d982d68df1adb37623698b2e which selects the entire first line of an element:
screen shot 2016-08-11 at 4 21 53 pm
@ghillairet please review update my fix and update if needed.

tedepstein: Very glad to see that it's the property name line selected. That's very clean from UX perspective, takes the user exactly where he wanted to go.

com.reprezen.swagedit/src/com/reprezen/swagedit/model/NodeDeserializer.java
```diff
@@ -0,0 +1,84 @@
...
+
+                AbstractNode v = deserialize(p, context);
+                v.setProperty(name);
+                node.put(name, v);
+            }
+
+            node.setEndLocation(p.getCurrentLocation());
+            return node;
+
+        } else if (p.getCurrentToken() == JsonToken.START_ARRAY) {
```
tfesenko, line 50: What about extracting methods to handle `START_OBJECT`,  `START_ARRAY` and the default case instead of keeping them all inside a 70-line long method?

com.reprezen.swagedit/src/com/reprezen/swagedit/model/NodeDeserializer.java
```diff
@@ -0,0 +1,84 @@
...
+import com.fasterxml.jackson.databind.JsonDeserializer;
+
+public class NodeDeserializer extends JsonDeserializer<AbstractNode> {
+
+    @Override
+    public AbstractNode deserialize(JsonParser p, DeserializationContext context)
+            throws IOException, JsonProcessingException {
+
+        final Model model = (Model) context.getAttribute("model");
+        final AbstractNode parent = (AbstractNode) context.getAttribute("parent");
```
tfesenko, line 20: The magic strings `parent` and `pointer` are used at least three times. What about converting them into reusable constants?

com.reprezen.swagedit/src/com/reprezen/swagedit/model/AbstractNode.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private static SchemaDefinitionProvider provider = new SchemaDefinitionProvider();
+
+    private final JsonPointer pointer;
+    private String property;
+    private AbstractNode parent;
+    private JsonType type;
+    private JsonNode schema;
+
+    public final Set<SchemaDefinition> definitions;
```
tfesenko, line 29: `definitions` is public, but there is a corresponding `getDefinitions()`. Was the field made `public` by mistake?

com.reprezen.swagedit/src/com/reprezen/swagedit/model/AbstractNode.java
```diff
@@ -0,0 +1,150 @@
...
+    AbstractNode(AbstractNode parent, JsonPointer ptr) {
+        this(parent, ptr, null);
+    }
+
+    AbstractNode(AbstractNode parent, JsonPointer ptr, JsonLocation location) {
+        this.parent = parent;
+        this.pointer = ptr;
+        this.location = location;
+        this.definitions = provider.getDefinitions(pointer(ptr));
+        this.schema = definitions.isEmpty() ? null : Iterables.getFirst(definitions, null).definition;
```
tfesenko, line 46: `getFirst(definitions, null).definition` looks like a potential NPE

com.reprezen.swagedit/src/com/reprezen/swagedit/model/AbstractNode.java
```diff
@@ -0,0 +1,150 @@
...
+
+    public void setProperty(String name) {
+        this.property = name;
+    }
+
+    public Iterable<AbstractNode> elements() {
+        return Collections.emptyList();
+    }
+
+    public abstract String getText();
```
tfesenko, line 114: I suggest to move this abstract method next to other abstract methods defined above to make it easier to spot all of them at once

tfesenko, line 114: I just noticed that `ArrayNode` and `ObjectNode` have the same implementation of `getText()`:

``` java
return getProperty() == null ? "" : getProperty();
```

Only `ValueNode` has different implementation.
@ghillairet you can use it as a default `getText()` and override it for `ValueNode`

com.reprezen.swagedit/src/com/reprezen/swagedit/model/AbstractNode.java
```diff
@@ -0,0 +1,150 @@
...
+
+    public AbstractNode getParent() {
+        return parent;
+    }
+
+    public String getProperty() {
+        return property;
+    }
+
+    public void setProperty(String name) {
```
tfesenko, line 106: is it propertyName or propertyValue? Based on the parameter, it's more like `propertyName`, is it correct? Can we rename the method or parameter name then?

com.reprezen.swagedit/src/com/reprezen/swagedit/model/ArrayNode.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public Iterable<AbstractNode> elements() {
+        return elements;
+    }
+
+    @Override
+    public String getText() {
+        return getProperty() == null ? "" : getProperty();
```
tfesenko, line 47: What about 

``` java
Optional.<String>of(getProperty()).or("")
```

?

com.reprezen.swagedit/src/com/reprezen/swagedit/editor/outline/SwaggerContentOutlinePage.java
```diff
@@ -0,0 +1,66 @@
...
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.ui.part.IShowInTarget;
+import org.eclipse.ui.part.ShowInContext;
+import org.eclipse.ui.texteditor.IDocumentProvider;
+import org.eclipse.ui.views.contentoutline.ContentOutlinePage;
+
+import com.reprezen.swagedit.editor.SwaggerDocument;
+import com.reprezen.swagedit.model.Model;
+
+public class SwaggerContentOutlinePage extends ContentOutlinePage {
```
tfesenko, line 15: Nice

com.reprezen.swagedit/src/com/reprezen/swagedit/editor/outline/SwaggerContentOutlinePage.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public void createControl(Composite parent) {
+        super.createControl(parent);
+
+        TreeViewer viewer = getTreeViewer();
+        viewer.setContentProvider(new OutlineContentProvider());
+        viewer.setLabelProvider(new OutlineStyledLabelProvider());
+        viewer.addSelectionChangedListener(this);
+        viewer.setAutoExpandLevel(2);
```
tfesenko, line 41: I found a bug which may be related to `setAutoExpandLevel(2)`
I noticed that if I expand an element in the Outline View, then edit the element in the editor, the corresponding node in the outline which was previously expanded gets collapsed:
outline_collapse

It's a bug.

tfesenko:
## Code review

The overall structure is good, the code is well-organized. I left several suggestions about details and spotted a potential NPE.
Also, I suggest to add javadocs to the most important public methods of the new `model` package.

## QA
- Navigation does not work - fixed in   d90ea00
- Outline view nodes are collapsed to second level on model update, see https://github.com/RepreZen/SwagEdit/pull/178#discussion_r74504023. This should be fixed.

Merging this PR.
I will update outline icons in the next pull request. @ghillairet will be busy next week and will work more hours on the week of August 22. It's a straightforward change, I will just update them myself.

tfesenko:
Well, I was wrong about adding icons being an easy change.
The icons we use for SwagEdit describe meta-information, the YAML type of the property: 
`outline_document, outline_scalar, outline_mapping, outline_sequence, outline_mapping_scalar`.

The icons we use for RAPID-ML are specific types: method, resource, datatype, etc... 
So, it's not just replacing the icons, but also the logic in changing `com.reprezen.swagedit.editor.outline.OutlineStyledLabelProvider.getIcon(AbstractNode)` .

@tedepstein should I introduce the new icons now or should we wait for @ghillairet ?

tedepstein:
@tfesenko we should wait for @ghillairet .  I didn't mean for you to make this change yourself, just make sure Guillaume has access to the icon files, and any non-obvious information about what each one of them means.

I guessed that the logic for assigning an icon to a node would be similar in most respects to the logic, [described here](https://github.com/RepreZen/SwagEdit/issues/40#issuecomment-236367773), for assigning the type annotation suffix to a node. 

I'm also sure that we will have certain Swagger node types that don't have a corresponding icon type in our current icon set for RAPID-ML.  As those cases arise, we'll find a solution.

Since this isn't trivial to implement, you could open a separate issue for it.

com.reprezen.swagedit/src/com/reprezen/swagedit/editor/outline/SwaggerContentOutlinePage.java
ghillairet: PR for this bug https://github.com/RepreZen/SwagEdit/pull/185

tfesenko:
@ghillairet confirming that I can observe the same behavior on 7f6664671beadef4298933e2682f45a26e588e25

I did not notice that the cursor was placed just after the colon because the selection was empty and we were talking about selection during previous iterations. 
Can we take the best of two approaches by select the entire line and placing the cursor after the colon?

tedepstein:
@tfesenko , from what I can tell, that's not possible.  There is either a selected range of text, OR an insertion point.  

Eclipse editors might be a little unusual, because they display a flashing insertion point at the end of a selection range.  If the selection ends with a newline, the insertion point is displayed on the following line.  But as far as I can tell, this insertion point is _always_ at the end of the selection range; it's just there to show you where text will go if you start typing.

So I think the new selection behavior is correct, as shown in @ghillairet's screenshot.  It doesn't automatically select the property value, so if you were navigating to that property with the intent of changing the value, you have to do at least one more click or keyboard navigation. 

But it does clearly select the line, and that selection draws your eye immediately to the element you wanted to navigate to.  That's the intended benefit.

Issue #183, opened 2016-08-17
Schema validation doesn't work with YAML Aliases


tedepstein:
This problem was reported by a user in [support ticket #150](https://support.reprezen.com/helpdesk/tickets/150).  

YAML supports [anchors and aliases](http://yaml.org/spec/1.2/spec.html#id2760545), which allow nodes to be repeated in the YAML structure.  

This Swagger spec should be allowed, but fails validation:

```

---
swagger: "2.0"
info:
  version: 1.0.0
  title: Uber parameters

paths: {}

parameters:  

  law:
    name: law_applicability
    in: query
    description: scope of a law
    required: false
    type: string
    &scope_values enum: 
    - GLOBAL
    - REGIONAL
    - LOCAL

  brand:
    name: brand_reach
    in: query
    description: where a brand name is used
    type: string
    enum:
      *scope_values
```

The error is:

```
Failed to match exactly one schema:
 - nonBodyParameter:
    Failed to match exactly one schema:
     - queryParameterSubSchema:
         - value of type string is not allowed, value should be of type array
     - headerParameterSubSchema:
         - value query is not allowed, value should be one of "header"
         - value of type string is not allowed, value should be of type array
     - formDataParameterSubSchema:
         - value query is not allowed, value should be one of "formData"
         - value of type string is not allowed, value should be of type array
     - pathParameterSubSchema:
         - object has missing required properties "required"
         - value query is not allowed, value should be one of "path"
         - value of type string is not allowed, value should be of type array
 - bodyParameter:
     - object has properties "enum", "type" which are not allowed
     - value query is not allowed, value should be one of "body"
     - object has missing required properties "schema"
```

tfesenko:
A better usage of the aliases in this case would be 

``` yaml
swagger: "2.0"
info:
  version: 1.0.0
  title: Uber parameters

paths: {}

parameters:  

  law:
    name: law_applicability
    in: query
    description: scope of a law
    required: false
    type: string
    enum: &scope_values
    - GLOBAL
    - REGIONAL
    - LOCAL

  brand:
    name: brand_reach
    in: query
    description: where a brand name is used
    type: string
    enum: *scope_values
```

Online Swagger editor converts it to the following JSON:

``` json
{
    "swagger": "2.0",
    "info": {
        "version": "1.0.0",
        "title": "Uber parameters"
    },
    "paths": {},
    "parameters": {
        "law": {
            "name": "law_applicability",
[...]

PR #184, opened 2016-08-19
#183 Schema validation doesn't work with YAML Aliases

tfesenko:
Because Jackson parse does not handle YAML aliases well, we first process the JSON by snake-yaml which expands aliases, then we transform expanded YAML it to a `JsonNode` which is provided to JSON Schema validator. 

I have performance concerns about too many passes required for validation:
1. `parseYaml()` creates a `org.yaml.snakeyaml.nodes.Node`
2. `parseJson()` uses snake-yaml to create an `Object` (a `Map`)
3. The map is provided to Jackson to build a `JsonNode`
4. Then, JSON Schema validator traverses the `JsonNode` to see if it conforms to the Swagger JSON Schema
As discussed at today's SUM, these concerns can be addressed later.

@andylowry can you please review and try to merge ASAP?

@ghillairet can you please review too?

### DIFF
--- a/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/validation/ValidatorTest.xtend
+++ b/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/validation/ValidatorTest.xtend
@@ -17,6 +17,7 @@ import org.eclipse.core.resources.IMarker
 import org.junit.Test
 
 import static org.hamcrest.core.IsCollectionContaining.*
+import static org.hamcrest.CoreMatchers.*
 import static org.junit.Assert.*
 
 class ValidatorTest {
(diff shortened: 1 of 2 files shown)


tedepstein:
This one is time-sensitive and I can't review right now.  I know @andylowry and @daffinm are busy with knowledge transfer on multi-file support.  @tfesenko , maybe we should merge this, get it into API Studio for testing by our user, and complete code review next week, after your performance optimization (and hopefully with @ghillairet here to do the code review).

tedepstein:
Leaving this to you, @tfesenko

tfesenko:
Ok, merging.
We can always roll it back if needed.

PR #185, opened 2016-08-23
[#40] Implement Quick Outline

ghillairet:
Fix issue with expanded elements in the outline view. Updates made in the document now does not reset the state of the tree in the outline to default expanded elements.

quick


tfesenko:
Passed code review and QA. Looks good!

Issue #193, opened 2016-08-26
NPE when the Outline is closed


tfesenko:
Extracted from https://github.com/ModelSolv/RepreZen/pull/1039#issuecomment-242767094

An NPE is thrown when the Outline View is not open:
![ee85cc3c-6b7f-11e6-81e5-6349526cbed6](https://cloud.githubusercontent.com/assets/644582/18015531/16f72a74-6b96-11e6-8921-9a98441a9da0.png)

It corresponds to

``` java
contentOutline.setInput(getEditorInput());
```

@ghillairet I assume this is an easy fix. Can you please consider it a high-priority bug.

PR #194, opened 2016-08-29
[#193] NPE when the Outline is closed

ghillairet:


tfesenko:
Straightforward fix. Merging

Issue #189, opened 2016-08-23
Warning instead of Error on curly braces in path item references


tedepstein:
Noted in [support ticket #152](https://support.reprezen.com/helpdesk/tickets/152).  

While SwagEdit disallows path item references with un-encoded curly braces, the Swagger.io editor allows them, and Swagger-UI actually _requires_ them to be un-encoded.  It won't work with URL-encoded curly braces.

Swagger-Normalizer is our immediate solution to this problem.  But users will want to be able to use un-encoded curly braces for tool portability, consistency, and readability.

Related Swagger-JS issue: https://github.com/swagger-api/swagger-js/issues/837

ghillairet:
Possible solution in PR https://github.com/RepreZen/SwagEdit/pull/197

Previous behavior

![screen shot 2016-08-30 at 15 48 46](https://cloud.githubusercontent.com/assets/148045/18093805/7ede12b8-6ed1-11e6-9d39-207c3eb4cab3.png)

New behavior

![screen shot 2016-08-30 at 15 49 04](https://cloud.githubusercontent.com/assets/148045/18093837/97d4e7ec-6ed1-11e6-9c76-3e2a0323dea6.png)

PR #197, opened 2016-08-30
[#189] Warning instead of Error on curly braces in path item references

ghillairet:
#189

### DIFF
--- a/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/validation/ReferenceValidatorTest.xtend
+++ b/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/validation/ReferenceValidatorTest.xtend
@@ -77,9 +77,34 @@ class ReferenceValidatorTest {
 			new SwaggerError(9, IMarker.SEVERITY_WARNING, Messages.error_missing_reference)
 		))
 	}
-	
+
 	@Test
 	def void shouldValidateReference_To_ValidPath() {
+		val content = '''
+			swagger: '2.0'
+			info:
+			  version: 0.0.0
+			  title: Simple API
+			paths:
+			  /foo/{bar}:
+			    get:
+			      parameters:
+			        - $ref: '#/paths/~1foo~1%7Bbar%7D'
+			      responses:
+			        '200':
+			          description: OK
+		'''
+
+		document.set(content)
+		val baseURI = new URI(null, null, null)
+		val resolvedURI = new URI(null, null, "/paths/~1foo~1{bar}")
+		val errors = validator(#{resolvedURI -> document.asJson}).validate(baseURI, document.yaml)
+
+		assertEquals(0, errors.size())
+	}
+
+	@Test
+	def void shouldWarnOnInvalidCharacters() {
 		val content = '''
 			swagger: '2.0'
 			info:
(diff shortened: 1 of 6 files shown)


tfesenko:
The PR introduces the following regression:
using a `./` in the ref value causes a warning:
screen shot 2016-08-30 at 5 37 26 pm
Removing the `./` resolves it:
screen shot 2016-08-30 at 5 37 37 pm

`../` is also OK (no warnings). 
Models:
[refs.zip](https://github.com/RepreZen/SwagEdit/files/445968/refs.zip)

We use the `./`-s in our Multi-file Uber example.

I will re-review this PR again tomorrow with a fresh head.

tfesenko:
Curly braces are supported in both unencoded human-friendly form and encoded form:
screen shot 2016-09-02 at 6 01 45 pm
As expected (and as it was in previous versions), code-assist proposes the encoded string for ref values.
I used these models - [refs2.zip](https://github.com/RepreZen/SwagEdit/files/452986/refs2.zip) .

But because of newly introduced warnings for the `./` in `$refs`, it's really hard to work with this version.  Please fix it.

tfesenko:
[refs2.zip](https://github.com/RepreZen/SwagEdit/files/452986/refs2.zip)

ghillairet:
Updated to only warn on curly braces, not other invalid characters.

com.reprezen.swagedit.tests/src/com/reprezen/swagedit/validation/ReferenceValidatorTest.xtend
```diff
@@ -104,6 +104,32 @@ class ReferenceValidatorTest {
...
+			          description: OK
+		'''
+
+		document.set(content)
+		val baseURI = new URI(null, null, null)
+		val resolvedURI = new URI(null, null, "/paths/~1foo~1{bar}")
+		val errors = validator(#{resolvedURI -> document.asJson}).validate(baseURI, document.yaml)
+
+		assertEquals(1, errors.size())
+		assertTrue(errors.contains(new SwaggerError(9, IMarker.SEVERITY_WARNING, Messages.error_invalid_reference)))
```
tfesenko, line 129: Nice

com.reprezen.swagedit/src/com/reprezen/swagedit/json/references/JsonReference.java
```diff
@@ -36,16 +36,19 @@
...
     private final boolean local;
     private final Object source;
+    private final boolean containsWarning;
 
     private JsonNode resolved;
     private JsonDocumentManager manager = JsonDocumentManager.getInstance();
 
-    JsonReference(URI uri, JsonPointer pointer, boolean absolute, boolean local, Object source) {
+    JsonReference(URI uri, JsonPointer pointer, boolean absolute, boolean local, boolean containsWarning,
+            Object source) {
```
tfesenko, line 45: We already reached 6 constructor parameter. In three usages of four we use the same values for the first five params: `new JsonReference(null, null, false, false, false, node)`.
If we ever decide to add another constructor parameter it would be good to see how we can make it more readable, e.g. use fluent builder API or create new constructors (or static methods) for most commonly used cases.

tfesenko:
Passed QA and code review

Issue #199, opened 2016-09-08
QuickOutline filtering should be case-insensitive


tedepstein:
In the "Pet Store with Examples" project below, a search for "toto" in quick outline view shows an empty list; while a search for "Toto" shows the expected nodes.

![image](https://cloud.githubusercontent.com/assets/692813/18349694/8ad9d112-75a0-11e6-829e-4d23ea151e7e.png)

The search should be case-insensitive, so that the user can type quickly and find the desired target nodes.

```
swagger: "2.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
host: petstore.swagger.io
basePath: /v1
schemes:
  - http
consumes:
  - application/json
produces:
  - application/json
paths:
  /pets:
    get:
      summary: List all pets
      operationId: listPets
      tags:
        - pets
      parameters:
        - name: limit
          in: query
          description: How many items to return at one time (max 100)
          required: false
          type: integer
          format: int32
      responses:
        200:
          description: An paged array of pets
          headers:
            x-next:
              type: string
              description: A link to the next page of responses
          examples:
            application/json:
              - 
                id: 1
                tag: dog
                name: Toto
              - 
                id: 2
                tag: cat
                name: Garfield
              - 
                id: 3
                tag: dog
                name: Snoopy
              - 
                id: 4
                tag: rat
                name: Remy
              - 
                id: 5
                tag: lizard
                name: Rango
          schema:
            $ref: "#/definitions/Pets"
        default:
          description: unexpected error
          schema:
            $ref: "#/definitions/Error"
    post:
      summary: Create a pet
      operationId: createPets
      tags:
        - pets
      responses:
        201:
          description: New Pet record added successfully
          examples:
            application/json:
              id: 1
              tag: dog
              name: Toto

        default:
          description: unexpected error
          schema:
            $ref: "#/definitions/Error"
  /pets/{petId}:
    get:
      summary: Info for a specific pet
      operationId: showPetById
      tags:
        - pets
      parameters:
        - name: petId
          in: path
          required: true
          description: The id of the pet to retrieve
          type: string
      responses:
        200:
          description: Expected response to a valid request
          schema:
            $ref: "#/definitions/Pet"
          examples:
            application/json:
              id: 1
              name: Toto
              tag: dog
        default:
          description: unexpected error
          schema:
            $ref: "#/definitions/Error"
definitions:
  Pet:
    required:
      - id
      - name
    properties:
[...]

PR #201, opened 2016-09-12
[#199] QuickOutline filtering should be case-insensitive

ghillairet:
[#199] QuickOutline filtering should be case-insensitive


tfesenko:
Passed core review and QA:
screen shot 2016-09-12 at 2 57 10 pm

Merging

PR #208, opened 2016-09-26
ZEN-2914 - loosen requirements for YEdit dependency

tfesenko:
Current SwagEdit requires the most recent YEdit to be installed. As a
result, SwagEdit cannot be installed on top of installations that
require older versions of YEdit, e.g. http://www.nodeclipse.org/. It
requires 1.0.15, not older, not newer. 

I modified our dependency on YEdit, so we now accept YEdit 1.0.15+ .
I tested this installation manually - installed withput problems, code
assist, code templates, and validation work as expected.
The release notes for YEdit (https://github.com/oyse/yedit/releases)
don't contain any incompatible change  between versions 1.0.15 and
1.0.20.

When update site created by this PR is installed on top of Eclipse with  YEdit 1.0.15 the old version of Yedit is preserved: 
screen shot 2016-09-26 at 2 41 09 pm

### DIFF
--- a/com.reprezen.swagedit.feature/feature.xml
+++ b/com.reprezen.swagedit.feature/feature.xml
@@ -105,7 +105,7 @@ This Agreement is governed by the laws of the State of New York and the intellec
       <import plugin="org.eclipse.core.resources"/>
       <import plugin="org.eclipse.ui.ide"/>
       <import plugin="org.eclipse.ui.workbench.texteditor"/>
-      <import plugin="org.dadacoalition.yedit" version="1.0.20" match="greaterOrEqual"/>
+      <import plugin="org.dadacoalition.yedit" version="1.0.15" match="greaterOrEqual"/>
    </requires>
 
    <plugin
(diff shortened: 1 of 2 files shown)


tfesenko:

tedepstein:
Thanks for the explanation and the fix, @tfesenko .  Can we also contact the Nodeclipse team and ask them to relax their YEdit dependency, to allow for newer YEdit releases?

tedepstein:
... Also, should we open an issue to review our own dependencies, and make sure we're not unnecessarily requiring old versions of any plugins?

ghillairet:
@tfesenko Looks good

tfesenko:
@tedepstein wrote:

> Can we also contact the Nodeclipse team and ask them to relax their YEdit dependency, to allow for newer YEdit releases?

Sure. Submitted https://github.com/Nodeclipse/nodeclipse-1/issues/216

> Also, should we open an issue to review our own dependencies, and make sure we're not unnecessarily requiring old versions of any plugins?

Created https://modelsolv.atlassian.net/browse/ZEN-2952

tfesenko:
And I forgot to say the YEdit is the only SwagEdit dependency for which we specify version in `feature.xml`.

Issue #205, opened 2016-09-19
Add additional information to content assist


ghillairet:
Content assist could use the description values present in the swagger schema as additional information.

Additional information are shown in a separate modal next to content assist proposals.

![screen shot 2016-09-19 at 16 55 49](https://cloud.githubusercontent.com/assets/148045/18636825/76ca2850-7e8a-11e6-9fe6-df60ad8bb28a.png)

tedepstein:
@ghillairet , nice idea.  The screenshot makes it look like this is already done.  But I assume there is still some work remaining, or maybe this is just a mockup?

In any case, please advise as to how much effort you think this would need.

ghillairet:
It's already in the model branch (#198), there is not much work to be done to have it finalized, only make sure all elements that have a description in their schema, have that additional information added.

tedepstein:
> only make sure all elements that have a description in their schema, have that additional information added.

Not sure what this means.  If an element has a description in schema, shouldn't the implementation use that description dynamically?  It seems like you're saying we would need to extract that information to some other form, for presentation in the information  window.  But I would hope we can avoid that.

ghillairet:
What I meant is that what's left to do is make sure the code that takes the description from the schema does it correctly, e.g. it does not miss any. But anyway the way it is done it should not.

tedepstein:
OK, @ghillairet , thanks for clarifying.

tfesenko:
Fixed in PR #210:
<img width="679" alt="screen shot 2016-10-01 at 12 38 03 pm" src="https://cloud.githubusercontent.com/assets/644582/19015536/86dbdd4e-87d4-11e6-912f-8b991d320298.png">

Issue #198, opened 2016-09-06
SwagEdit Model


ghillairet:
This issue is to discuss all changes made to support a better representation of the content of a swagger document. Work on this is currently done in the branch task/model. 

The goals of this work is to unify the access to swagger document content, such as position and types of elements, in the different services provided by SwagEdit such as content assist, outlines, etc.. This should simplify a lot of code that now uses either the Jackson Node API or snakeyaml API to achieve this.

> Background, this work was started with quick outline, see https://github.com/RepreZen/SwagEdit/pull/176

List of tasks to be done
- [x] Implement Model parser with Jackson streaming API
- [x] Remove SchemaDefinition class, replaced by TypeDefinition and SwaggerSchema
- [x] Type definitions should be derived from swagger JSON schema
- [x] Content assist proposals should be obtained through Node type
- [x] Content assist processor should use the new Model API
- [x] JSON pointers should be used instead of previous path format to navigate inside a document
- [x] Outline label provider should use the Model API
- [x] Fix method SwaggerDocument.getRegion(node) that often returns incorrect location
- [x] Make sure tests work with maven/tycho
- [x] All tests are not yet correct, some of them are still failling
- [x] Looks like validation takes more time than before, this should be fixed

ghillairet:
Current status on this.

The hyperlinks detectors and the content assist processor have been updated to start using the new model API.

Work on the resolution of types from the swagger schema has been started. It was previously done by using the SchemaDefinition class, which has been removed. Resolution was happening each time a call to a content assist was performed. Now all types are resolved when the editor starts and are stored and accessed by the class SwaggerSchema.

Moving all paths to JSON pointers has also been started and is yet to be completed.

tfesenko:
Notes for @tfesenko for next week's review:

## Verify

That aliases still work with the new code

## Suggestions:
- Merge conflicts - task branch cannot be merged into master
- `com.reprezen.swagedit.json.TypeDefinition.resolve(JsonNode, JsonNode)` -> `resolveRef()`
- `com.reprezen.swagedit.json.JsonUtil.asPointer(String)` and `getPointer()` look similar. Can `getPointer()` call `asPointer()`? Why does `asPointer()` sanitize the string and `getPointer` does not?
[...]

Issue #58, opened 2016-01-07
Code assist should support typeahead with dynamic filtering


tedepstein:
The code assist feature in other Eclipse editors supports two useful behaviors that we should add to SwagEdit: 
1.  After invoking code assist, you can start typing, and the list of suggestions will remain visible, filtered to show the suggestions that start with the substring you entered.
2. If you invoke code assist with one or more characters already typed, immediately to the left of the insertion point, and then invoke code assist, the suggestion list will be pre-filtered to show matching suggestions.

tedepstein:
I noticed that filtering works in some contexts, but not others.  It appears to work fine on a method.  Typing a 'd' filters down to the keywords "deprecated" and "description".  But it doesn't work on a parameter item.  Typing a letter just closes the code assist suggestions, instead of filtering it.

tedepstein:
Making this one high priority.  Some users are starting to notice that this doesn't work as expected, and in some cases they have the impression that code assist simply doesn't work.

ghillairet:
I have looked at this. The problem is that in some cases, the yaml parser fails to parse the content of the document after the user enters a character and invoke content assist. The document becomes invalid while the user enters those new characters.
We need what is returned by the yaml parser to compute the content assist proposals, but in those cases maybe we should allow the proposals provider to use the previous state of the document.

tfesenko:
As @ghillairet mentioned above this task is more difficult than just filtering a list of strings based on a prefix. We have to deal with a parser results for invalid models. For example, Xtext generates a special lightweight Antlr parser (`org.eclipse.xtext.ui.editor.contentassist.antlr.internal.AbstractInternalContentAssistParser`) and lexer tailored for content assist.
Finding a right approach for implementation of this feature can take some time.

tedepstein:
@tfesenko @ghillairet, thanks for the input.  I continue to think that this is really important, and we should try to find a way to make it work.  I'm quite open to considering some creative solutions for this, if you have any.

tedepstein:
Not sure if @ghillairet's most recent commit, [introducing a unified JSON model](https://github.com/RepreZen/SwagEdit/commit/e0d612d802e43792b118a4287581b08d25d23897), helps with typeahead.  But I had a couple of thoughts about this that I wanted to record here, when we're ready to get back to it:
[...]

PR #210, opened 2016-09-27
Task/model

ghillairet:
PR for #198 

This also includes enhancements for content assist #205 and preliminary enhancements for #58


tfesenko:
QA:
 I found only one regression so far:
Navigation from a path parameter to a parameter definition defined in another file does not work. To reproduce:
1. Extract the attached archive - [Archive.zip](https://github.com/RepreZen/SwagEdit/files/504466/Archive.zip)
2. Open `uber_path_estimates_time.yaml` line 26. Try opening a declaration (Ctrl+Click) - nothing happens. It worked in previous version.
Code assist and validation for this case work correctly.

Extracted to #211

tfesenko:
Code review:
I already did initial review in  #198 . Since my last comment, the `json` packages was renamed to `schema`, some methods got better names and many classes got javadocs.
The code looks cleaner and better organized. 

Merging.

tfesenko:
@ghillairet I see that the following TODO items are still marked as incomplete:
- Outline label provider should use the Model API
- Fix method SwaggerDocument.getRegion(node) that often returns incorrect location
- Looks like validation takes more time than before, this should be fixed

Can you please create issues for them if needed? 
After that I will close this issue.

Issue #211, opened 2016-10-01
Navigation to externally defined parameters


tfesenko:
A regression extracted from https://github.com/RepreZen/SwagEdit/pull/210#issuecomment-250920983

Navigation from a path parameter to a parameter definition defined in another file does not work. To reproduce:
1. Extract the attached archive - Archive.zip
2. Open uber_path_estimates_time.yaml line 26. Try opening a declaration (Ctrl+Click) - nothing happens. It worked in previous version.
Code assist and validation for this case work correctly.

PR #212, opened 2016-10-04
[#211] Navigation to externally defined parameters

ghillairet:
Fixes the issue with hyperlinks to external references due to bad resolution of nodes from a location inside the document. See #211 

This PR also includes missing Copyright headers.

### DIFF
--- a/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/editor/SwaggerDocumentTest.xtend
+++ b/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/editor/SwaggerDocumentTest.xtend
@@ -217,4 +217,22 @@ class SwaggerDocumentTest {
 		test.apply("/paths/~1testing/get/parameters/1", "2")
 	}
 
+	@Test
+	def void testPointerParameter() {
+		val doc = new SwaggerDocument
+		val test = Cursors.setUpPathTest('''
+			paths:
+			  /pets:
+			    get:
+			      parameters:
+			        - $r<1>ef: 's<2>ome_ref'
+			        - $<3>ref: 'some_o<4>ther_ref'
+		''', doc)
+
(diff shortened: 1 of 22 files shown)


com.reprezen.swagedit.tests/src/com/reprezen/swagedit/editor/SwaggerDocumentTest.xtend
```diff
@@ -217,4 +217,22 @@ class SwaggerDocumentTest {
...
+		val test = Cursors.setUpPathTest('''
+			paths:
+			  /pets:
+			    get:
+			      parameters:
+			        - $r<1>ef: 's<2>ome_ref'
+			        - $<3>ref: 'some_o<4>ther_ref'
+		''', doc)
+
+		test.apply("/paths/~1pets/get/parameters/0", "1")
```
tfesenko, line 232: Cool!
So here we verify that at `<1>` we are getting a path `/paths/~1pets/get/parameters/0`. 
I really like the idea, but it needs some mental work and reading the sources to understand what exactly `test.apply` is doing. It's not urgent, but it would be helpful to rename classes and/or methods or at least add an explanatory comment to `Cursors`

com.reprezen.swagedit/src/com/reprezen/swagedit/model/Model.java
```diff
@@ -133,23 +142,29 @@ public AbstractNode getNode(int line, int column) {
...
+                    }
                 }
             }
         }
         return current;
     }
 
     protected AbstractNode findCorrectNode(AbstractNode current, int column) {
-        if (startColumn(current) == column) {
+        if (current.getStart().getColumn() == column) {
```
tfesenko, line 161: Nice

tfesenko (approved):

tfesenko:
Passed QA and code review

Issue #215, opened 2016-10-04
Code-assist: `file` as value for responses' `type`


tfesenko:
Code-assist suggests `file` as value for responses' `type`
<img width="391" alt="screen shot 2016-10-04 at 4 50 35 pm" src="https://cloud.githubusercontent.com/assets/644582/19092732/07b12f68-8a56-11e6-89e5-3ab883dab512.png">

I expect to see`string`, `boolean`, and other values available for request's type:
<img width="263" alt="screen shot 2016-10-04 at 5 15 51 pm" src="https://cloud.githubusercontent.com/assets/644582/19092785/507c5f92-8a56-11e6-8a33-96ba7ccec9e4.png">

Model (cursor is at `<1>`)

```
swagger: "2.0"
info:
  title: TEST API
  description: Test
  version: "1.0.0"
host: api.uber.com
schemes:
  - https
basePath: /v1
produces:
  - application/json
  - application/xml
paths:
  /products:
    get:
      summary: Product Types
      description: some desc
      parameters:
        - name: prop
          in: header
          description: Latitude component of location.
          required: true
          type: string
          format: double
      responses:
        200:
          description: foo
          schema:
            type: <1>
```

ghillairet:
Those are different, there is only one value possible for the `type` property in a response schema, see https://github.com/RepreZen/SwagEdit/blob/master/com.reprezen.swagedit/src/com/reprezen/swagedit/schema/schema.json#L1074

tedepstein:
@ghillairet , [Looking at the usage](https://github.com/RepreZen/SwagEdit/blob/master/com.reprezen.swagedit/src/com/reprezen/swagedit/schema/schema.json#L361) of `fileSchema` in `response`, we can see that it is `oneOf` two possible types, the other being `schema`, which has all of those other possible values for `type`.

Ideally, where there is a `oneOf` or `anyOf` construct like this, code assist should offer a union of the available values from the sub-schemas.  

Have we considered this, and are we doing this in other `oneOf` or `anyOf` cases?

ghillairet:
Yes that's right, I overlooked this. We do union of values in that case, it does not work here because we yet don't get values from the json core schema (in that case https://github.com/RepreZen/SwagEdit/blob/master/com.reprezen.swagedit/src/com/reprezen/swagedit/schema/schema.json#L1001).

tedepstein:
[...]

PR #221, opened 2016-10-11
[#215] Code-assist: `file` as value for responses' `type`

ghillairet:
This commit adds support for the core json schema, and resolves the issue with missing content assist information that is available only from that core schema. See #215

### DIFF
--- a/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/assist/SwaggerContentAssistProcessorTest.xtend
+++ b/com.reprezen.swagedit.tests/src/com/reprezen/swagedit/assist/SwaggerContentAssistProcessorTest.xtend
@@ -170,18 +170,97 @@ class SwaggerContentAssistProcessorTest {
 	def void test3() {
 		val test = setUpContentAssistTest('''
 		 definitions:
+		  Foo:
+		    <1>
 		  Product:
-		    <1>required:
+		    <2>required:
 		      - name  
 		    properties:
 		      name:
 		        type: string
 		      description:
-		        type: <2>
+		        type: <3>
 		''', new SwaggerDocument)
-		
+
 		var proposals = test.apply(processor, "1")
+		var values = proposals.map[(it as StyledCompletionProposal).replacementString]
+		assertEquals(30, values.size)
+		assertThat(values, hasItems(
+			"$ref:", 
+			"format:", 
+			"title:", 
+			"description:", 
+			"multipleOf:", 
+			"maximum:", 
+			"exclusiveMaximum:", 
+			"minimum:", 
+			"exclusiveMinimum:", 
+			"maxLength:", 
+			"minLength:", 
+			"pattern:", 
+			"maxItems:", 
+			"minItems:", 
+			"uniqueItems:", 
+			"maxProperties:",
+			"minProperties:", 
+			"required:", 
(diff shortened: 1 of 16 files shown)


com.reprezen.swagedit/src/com/reprezen/swagedit/schema/JsonType.java
```diff
@@ -59,6 +59,8 @@ public static JsonType valueOf(JsonNode node) {
             return JsonType.OBJECT;
         } else if (node.has("anyOf")) {
             return JsonType.ANY_OF;
+        } else if (node.has("allOf")) {
+            return JsonType.ALL_OF;
```
tfesenko, line 63: Thanks!
Can we group checks for anyOf, allOf and **oneOf** together?

com.reprezen.swagedit.tests/src/com/reprezen/swagedit/assist/SwaggerContentAssistProcessorTest.xtend
```diff
@@ -170,18 +170,97 @@ class SwaggerContentAssistProcessorTest {
...
+			"readOnly:", 
+			"xml:", 
+			"externalDocs:", 
+			"example:", 
+			"x-:"))
+
 		proposals = test.apply(processor, "2")
-//		println(proposals.map[(it as StyledCompletionProposal).replacementString])
+		values = proposals.map[(it as StyledCompletionProposal).replacementString]
+		// same without required and properties
```
tfesenko, line 222: Thank you for the comment, I was about to ask why are we having 28 entries here instead of 30.

com.reprezen.swagedit.tests/src/com/reprezen/swagedit/assist/SwaggerProposalProviderTest.xtend
```diff
@@ -203,18 +203,40 @@ class SwaggerProposalProviderTest {
...
+		assertEquals(8, values.size)
+		assertThat(values, hasItems(
+			"array",
+			"boolean",
+			"integer",
+			"null",
+			"number",
+			"object",
+			"string",
+			"file"
```
tfesenko, line 239: Thanks!

tfesenko:
## Code review

Passed code review. Changes overview:
- ReferenceTypeDefinition - new type
- SwaggerSchema.JsonSchema - new type, incorporates some information which was formerly defined in SwaggerSchema and TypeDefinition.create()
- Proposal - change hashCode and equals()
- SwaggerProposalProvider.getProposals(.., prefix)
- Tests

## QA

Simple types are now shown by code assist:
screen shot 2016-10-13 at 3 58 35 pm
Note that now we have a `null` as a proposal. It's taked from core.json schema:

``` json
        "simpleTypes": {
            "enum": [ "array", "boolean", "integer", "null", "number", "object", "string" ]
        },
```

But selected it causes a validation error: 
screen shot 2016-10-13 at 4 08 11 pm

@ghillairet , can you please investigate why it's happening? 

> and should we aim to do this in a way that doesn't require an internet connection

Works offline.

ghillairet:
The issue with `null` values has been fixed by adding quotes around it. Without quotes the value is parsed as null (no values) and not as the "null" string as it should be for the validation to work.

Issue #202, opened 2016-09-13
Multi-File QuickOutline View


tedepstein:
After we fix code assist typeahead, next priority for us is multi-file support in Quick Outline View.

This should cycle through current file, project files and workspace files, the same as multi-file code assist.

To clarify the intent:  this is not about expanding references in QuickOutline view.  What we want to do is introduce top-level nodes for Swagger files, and populate the outline under each of these files.  The filter/search feature should filter across the entire outline, and this should allow fast access to any node within any swagger file in the workspace.

ghillairet:
Preliminary implementation for this has been started in PR #225 

Screencast showing it working

![out1](https://cloud.githubusercontent.com/assets/148045/19442513/d3ddd482-9489-11e6-88ea-f09505b5f576.gif)

The content of the quick outline is being updated when the user presses the same keys used to open the quick outline (e.g. Ctrl-o).

The status of the quick outline (at the bottom of the popup) is also being updated with the following messages (where %s is replaced with the keys):
- Press '%s' to show elements only the from current file
- Press '%s' to show elements from the current project files
- Press '%s' to show elements from the current workspace files

What remain to do is fix navigation to files other than current (when user clicks on tree element) and improve performance of text filter (which hangs on large and/or multiple files).

PR #225, opened 2016-10-17
[#202] Multi-File QuickOutline View

ghillairet:
See issue #202

Preliminary PR for multi file support in quick outline, should not be merged now, still issues regarding navigation and performance of text filter.

### DIFF
--- a/com.reprezen.swagedit/src/com/reprezen/swagedit/editor/outline/QuickOutline.java
+++ b/com.reprezen.swagedit/src/com/reprezen/swagedit/editor/outline/QuickOutline.java
@@ -10,6 +10,10 @@
  *******************************************************************************/
 package com.reprezen.swagedit.editor.outline;
 
+import org.eclipse.core.resources.IFile;
+import org.eclipse.jface.bindings.TriggerSequence;
+import org.eclipse.jface.bindings.keys.KeySequence;
+import org.eclipse.jface.bindings.keys.SWTKeySupport;
 import org.eclipse.jface.dialogs.Dialog;
 import org.eclipse.jface.dialogs.PopupDialog;
 import org.eclipse.jface.text.IInformationControl;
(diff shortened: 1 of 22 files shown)


com.reprezen.swagedit/src/com/reprezen/swagedit/utils/SwaggerFileFinder.java
```diff
@@ -0,0 +1,128 @@
...
+import org.eclipse.core.resources.IResourceProxy;
+import org.eclipse.core.resources.IResourceProxyVisitor;
+import org.eclipse.core.runtime.CoreException;
+
+import com.google.common.collect.Lists;
+
+public class SwaggerFileFinder {
+
+    /**
+     * Represents the scope for which the JSON reference proposals have to be computed. <br/>
```
tfesenko: `Scope` was moved from `JsonReferenceProposalProvider` and now serves several purposes,  it's not limited to reference proposals anymore. The javadoc should be updated accordingly.

com.reprezen.swagedit/src/com/reprezen/swagedit/utils/SwaggerFileFinder.java
```diff
@@ -0,0 +1,128 @@
...
+
+        try {
+            parent.accept(visitor, 0);
+        } catch (CoreException e) {
+            return Lists.newArrayList();
+        }
+        return visitor.getFiles();
+    }
+
+    public Iterable<IFile> collectFiles(Scope scope, IFile currentFile) {
```
tfesenko, line 88: Nice

com.reprezen.swagedit/src/com/reprezen/swagedit/editor/outline/QuickOutline.java
```diff
@@ -110,6 +135,48 @@ public void modifyText(ModifyEvent e) {
...
+        if (input instanceof IFileEditorInput) {
+            currentFile = ((IFileEditorInput) input).getFile();
+        }
+
+        Iterable<IFile> files = fileFinder.collectFiles(currentScope, currentFile);
+        setInfoText(statusMessage());
+        if (currentScope == Scope.LOCAL) {
+            treeViewer.setAutoExpandLevel(2);
+        } else {
+            treeViewer.setAutoExpandLevel(0);
```
tfesenko, line 157: Cool

com.reprezen.swagedit/src/com/reprezen/swagedit/model/Model.java
```diff
@@ -84,6 +97,33 @@ public static Model parseYaml(SwaggerSchema schema, String text) {
...
+        final List<Model> modelSet = Lists.newArrayList();
+        for (IFile file : files) {
+            Model model = new Model(schema, file.getFullPath());
+            try {
+                mapper.reader() //
+                        .withAttribute(ATTRIBUTE_MODEL, model) //
+                        .withAttribute(ATTRIBUTE_PARENT, null) //
+                        .withAttribute(ATTRIBUTE_POINTER, JsonPointer.compile("")) //
+                        .withType(AbstractNode.class) //
+                        .readValue(file.getLocationURI().toURL());
```
tfesenko: `ObjectReader` we construct here is identical to the one created in`parseYaml(SwaggerSchema schema, String text)`. Can we reuse the code?

tfesenko (commented):
Changes overview
Outline:
`QuickOutline` implements new functionality, check `handleMultiView() and`isInvocationEvent`
`OutlineStyledLabelProvider` - defines new labels which may include path.

Related changes:
- `AbstractNode` has a `Model`
- `SwaggerFileFinder` to traverse files, 
  JsonReferenceProposalProvider$FileVisitor, $Scope, #collectFiles() -> SwaggerFileFinder
- a new overload for collectFiles() with a scope as a parameter
- Model now has `path` parameter

ghillairet:
Navigation is now working as expected. It should be ready for testing and merge if ok. One thing that could be improved if necessary is performance of filter in case large and/or multiple models being display in the quick outline.

tedepstein:
> One thing that could be improved if necessary is performance of filter in case large and/or multiple models

@tfesenko , suggest you test this with one of Saul's RopeSwagger multi-file models.  You can find these as attachments to related FreshDesk support issues.

com.reprezen.swagedit/src/com/reprezen/swagedit/model/NodeDeserializer.java
```diff
@@ -107,11 +105,10 @@ protected ValueNode deserializeValueNode(JsonParser p, DeserializationContext co
 
         Object v = context.readValue(p, Object.class);
 
-        ValueNode node = new ValueNode(model, parent, ptr, v);
+        ValueNode node = model.valueNode(parent, ptr, v);
```
tfesenko, line 108: I really like the new `objectNode()`, `valueNode` and `arrayNode()` methods. Because they are public we can communicate their intention better by renaming them to something like `addObjectNode()` or `createObjectNode`

com.reprezen.swagedit/src/com/reprezen/swagedit/utils/DocumentUtils.java
```diff
@@ -104,10 +124,49 @@ public static IFile getWorkspaceFile(IPath path) {
...
+        }
+    }
+
+    /**
+     * Opens the editor for the file located at the given path and reveal the selection.
+     * 
+     * @param path
+     * @param selection
+     */
+    public static void openAndReveal(IPath path, ISelection selection) {
```
tfesenko, line 153: So basically we expect that `selection` will wrap an `AbstractNode`, right?

tfesenko (approved):

tfesenko:
Works as expected - the different scope work as expected, navigation works for both local and multi-file modes.

I did not notice any performance issues on Saul's models using my laptop with 8 GB RAM while manually expanding elements or working using a local scope. 
But when I filter for an element in project scope, there is a visible delay. Here I am starting with a local scope, then use the workspace scope. A stopwatch is placed on the right side for your convenience:
quick_outline

## Some screenshots

Local scope:
screen shot 2016-10-18 at 14 45 27

Project scope:
screen shot 2016-10-18 at 14 47 49

Workspace scope, as I have many projects, it shows a scroll bar:
screen shot 2016-10-18 at 14 45 40

We auto-expand local scope up to second level, and don't expand the tree at all in project and 
workspace levels. I think it would be more practical to auto-expand the local scope up to third level as this is where all "interesting" elements are - paths, definitions, etc:
screen shot 2016-10-18 at 14 51 43

Issue #233, opened 2016-10-24
Add features section to Readme


tedepstein:
We have some nice annotated graphics that we can use for a features section in the Readme.

PR #235, opened 2016-10-24
#233 - Add some feature highlight graphics to readme.

tedepstein:
We should eventually add others, just wanted to start down this road by reusing the graphics I had put together for the API Studio QRC.


tfesenko:
The new feature highlights look good!

Issue #228, opened 2016-10-18
Fix performance issues with Multi-File QuickOutline View 


tfesenko:
Extracted from https://github.com/RepreZen/SwagEdit/pull/225#issuecomment-254605459

tedepstein:
@tfesenko and @ghillairet , please have a look at this and plan some basic improvements.  Go for the 80% in the 80-20 rule.

tfesenko:
Enabling "Detect periods of unresponsive UI" on `General>UI Responsiveness Monitoring` preferences page is helpful to see stacktraces of long operations. 
<img width="928" alt="screen shot 2016-10-27 at 4 51 14 pm" src="https://cloud.githubusercontent.com/assets/644582/19784977/b6caaa58-9c65-11e6-8f1b-e3df40b9f9bb.png">

tfesenko:
Auto-expanding all elements for the project scope takes up to three seconds.
1. Here, I am opening the outline in the project scope. Its nodes are **not** expanded.
2. Then, I start filtering element by typing "q", then "w". It's slow, but below a second.
3. When I remove the last letter from my filter the Outline becomes really slow, it freezes for three seconds. After the freeze all its nodes are expanded. They should be collapsed, the same as in the step 1. 
![expand_all](https://cloud.githubusercontent.com/assets/644582/19786520/67b6e3f8-9c6c-11e6-84d2-5aa9c4578462.gif)

tfesenko:
## `AbstractTreeViewer.expandAll()` - the ultimate fix

`treeViewer.expandAll()` is where all freezes happen according to UI Responsiveness Monitoring.
Making it fast would be an ultimate fix. I don't know if it's possible. The fix may also take too much time.

### `tree.setRedraw()` (did not help)

I tried wrapping the contents of the `modifyText()` inside a `treeViewer.getTree().setRedraw(false)`/`treeViewer.getTree().setRedraw(true)`, but it did not help. It even make it slower.

## Other optimizations

### Auto-expanding elements in project scope with a an empty filter

Auto-expanding all elements for the project scope takes up to three seconds.
1. Here, I am opening the outline in the project scope. Its nodes are **not** expanded.
2. Then, I start filtering element by typing "q", then "w". It's slow, but below a second.
3. When I remove the last letter from my filter the Outline becomes really slow, it freezes for three seconds. After the freeze all its nodes are expanded. They should be collapsed, the same as in the step 1. 
[...]

PR #242, opened 2016-11-01
[#228] Fix performance issues with Multi-File QuickOutline View

ghillairet:
Add timer inside the filter's modify listener so that refresh only happens after a certain amount of time (500ms), avoiding refresh at each text input.

see issue #228

### DIFF
--- a/com.reprezen.swagedit/src/com/reprezen/swagedit/editor/outline/QuickOutline.java
+++ b/com.reprezen.swagedit/src/com/reprezen/swagedit/editor/outline/QuickOutline.java
@@ -10,6 +10,9 @@
  *******************************************************************************/
 package com.reprezen.swagedit.editor.outline;
 
+import java.util.Timer;
+import java.util.TimerTask;
+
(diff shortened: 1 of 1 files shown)


com.reprezen.swagedit/src/com/reprezen/swagedit/editor/outline/QuickOutline.java
```diff
@@ -121,20 +125,43 @@ public void keyReleased(KeyEvent e) {
...
+                filterTimer = new Timer();
+                filterTimer.schedule(new TimerTask() {
+                    @Override
+                    public void run() {
+                        if (filterText.isDisposed()) {
+                            return;
+                        }
+
+                        // Make sure we access the text in the correct thread.
+                        filterText.getDisplay().asyncExec(new Runnable() {
```
tfesenko, line 143: Why do we need an `asyncExec()`, why not a `syncExec()`?

tfesenko (approved):
Great usability improvement!
We can discuss the sync/async choice the next week.

PR #243, opened 2016-11-02
ZEN-2180 - customize SwagEdit default font size

tfesenko:
@daffinm , you asked me to provide a way to customize SwagEdit default font size which should not influence other editors. I added a SwagEdit font theme, it's available both in RCP and add-on modes:
screen shot 2016-11-02 at 4 21 33 pm
It applies Arial-regular-12 to Swagger editors by default. The default value can be changes in preferences.
I did not find Consolas in my installation and replaced it with Arial. Please let me know if you have better suggestions.

Before this change:
swagedit_oldfont

With this change:
swagedit_newfont

### DIFF
--- a/com.reprezen.swagedit/plugin.xml
+++ b/com.reprezen.swagedit/plugin.xml
@@ -210,9 +210,9 @@
         <themeElementCategory label="SwagEdit" id="com.reprezen.swagedit.editor.presentation"/>
         <fontDefinition
               categoryId="com.reprezen.swagedit.editor.presentation"
+              defaultsTo="org.eclipse.jface.textfont"
               id="com.reprezen.swagedit.editor.textfont"
-              label="SwagEdit Text Font"
-              value="Arial-regular-12">
+              label="SwagEdit Text Font">
          <description>
             This font is used by Swagger editor.
          </description>
--- a/com.reprezen.swagedit/plugin.xml
+++ b/com.reprezen.swagedit/plugin.xml
@@ -25,7 +25,8 @@
             default="false"
             icon="icons/swagger_16.png"
             id="com.reprezen.swagedit.editor"
-            name="%editor.name">
+            name="%editor.name"
+            symbolicFontName="com.reprezen.swagedit.editor.textfont">
          <contentTypeBinding
                contentTypeId="com.reprezen.swagedit.contenttype.swagger.yaml">
          </contentTypeBinding>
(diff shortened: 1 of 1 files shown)


tedepstein:
Will users be annoyed that they font is not the one they've set as their standard text font, and that they have to change it separately?

tfesenko:
@tedepstein  wrote:

> Will users be annoyed that they font is not the one they've set as their standard text font, and that they have to change it separately?

That's a good question. They will need to change it separately. There is a button "use system font" which sets the **system** font, but may not be the same as JFace text font (Basic>Text Font)... 

Setting  SwagEdit font independently from other fonts was a requirement from @daffinm . 
I can also keep the font category and set its default value to the value of Basic>Text Font (`defaultsTo="org.eclipse.jface.textfont"`) so they will be synchronized by default and the SwagEdit value could be modified if needed.

daffinm:
@tfesenko I didn't know my question about being able to set the SwagEdit font separately had turned into a requirement. I thought you said you would look into it and let me know if what I wanted was a) possible and b) how much effort it would take.

I am writing a KB article on how to make API Studio look pretty in add-on mode. This means the following things in order of precedence: 
1. Setting the color theme
2. FIxing highlighting on template expansion (if dark theme is chosen)
3. Adjust the font size for SwagEdit (RAPID editor handles this for itself)
4. Turn off current line highlighting (if dark color theme and edit box are used)

The current method of changing the font face and size affects things globally. I was asking if we could localise these effects to just SwagEdit somehow - via a property that could be updated via a preferences import and later via the preferences dialog.

Sorry if I did not make this as clear as I should have done.

daffinm:
HI @tfesenko. Thanks for explaining during SUM. To recap, once you had done the research, the quickest way of documenting what you had found out was to just do it: add the extension point enabling SwagEdit font to be specifically adjusted. Right now, the only way of adjusting this is via the `org.eclipse.jface.textfont` preference which affects all fonts used in Eclipse. This is what we use in the RCP plugin_customisation.ini file but we cannot use this in add-on mode. Hence the question. 

So, as long as the default value is the same as `org.eclipse.jface.textfont` then I am happy for this to be merged. Thanks.

tfesenko:
Set default SwagEdit fount to `org.eclipse.jface.textfont`:
screen shot 2016-11-03 at 2 12 24 pm
I am merging it now to create a testable build for @daffinm 
@ghillairet , please review the changes from this PR when you have a moment.

Issue #239, opened 2016-10-27
Swagger Spec pasted into editor shows spurious warnings until saved


tedepstein:
The attached Swagger Spec is fine, and shouldn't show any warnings.  But if you copy the text content and paste it into SwagEdit, all of the $ref values (and possibly some others) show warning markers.  

![image](https://cloud.githubusercontent.com/assets/692813/19771362/ad06e5de-9c30-11e6-944f-7859c891e4fa.png)

Saving the Swagger spec seems to resolve this.

This is not high-priority, but if there's a way to prevent this, it would be helpful.

[RestAPI.zip](https://github.com/RepreZen/SwagEdit/files/556142/RestAPI.zip)

ghillairet:
I cannot reproduce this issue with the version in the master. @tfesenko can you try to reproduce?

tfesenko:
I am not seeing it either. Tried both from sources (on master) and on RepreZen daily build (1.3.3.5609). Added edits to force-validate the model.

I think it was fixed with changed introduced last week.

tedepstein:
I am still able to reproduce the error on build 5605, from Friday.  Can you please try on Windows?

tfesenko:
I was able to reproduce it on a Mac. 
A key condition here is to **paste** the model text into an existing file which is open in SwaggerEditor. Immediately after pasting the model the editor shows these warnings for refs. After editing it (I added three whitespaces to an arbitrary place) the warnings disappear.

tedepstein:
@tfesenko , correct.  I said that in the issue title and description, but next time I'll put it in **bold**.  :-)

ghillairet:
Could be related to #206

ghillairet:
PR #246 should solve this issue, not sure that solves issues with #206 as it only solves issues with references validation.

PR #246, opened 2016-11-08
[#239] Swagger Spec pasted into editor shows spurious warnings until …

ghillairet:
…saved

Resolution of JSON references will be done, if references are locals, by inspecting the content of the current documents and not by loading the content of the file on disk. This will ensure that validation of references will be correct when new content is copy/paste inside the current editor.

See issue #239


tfesenko (approved):
Passed code review and QA.
We don't have warnings, but we have new errors because missing schema types :)

PR #256, opened 2017-01-10
Eliminate LinkageError when passing a JsonPointer loaded by a different plugin

tfesenko:
Also see PR #255

Because Jackson is shipped as a lib jar, external plugin fail to call
Model#find(JsonPointer):
 
```
java.lang.LinkageError: loader constraint violation: when resolving
method
"com.reprezen.swagedit.model.Model.find(Lcom/fasterxml/jackson/core/JsonPointer;)Lcom/reprezen/swagedit/model/AbstractNode;"
the class loader (instance of
org/eclipse/osgi/internal/loader/EquinoxClassLoader) of the current
class, and the class loader (instance of
org/eclipse/osgi/internal/loader/EquinoxClassLoader) for the method's
defining class, com/reprezen/swagedit/model/Model, have different Class
objects for the type com/fasterxml/jackson/core/JsonPointer used in the
signature
```

### DIFF
--- a/com.reprezen.swagedit/src/com/reprezen/swagedit/model/Model.java
+++ b/com.reprezen.swagedit/src/com/reprezen/swagedit/model/Model.java
@@ -200,6 +200,10 @@ public class Model {
         return nodes.get(pointer);
     }
 
+    public AbstractNode find(String pointer) {
+        return nodes.get(JsonPointer.valueOf(pointer));
+    }
+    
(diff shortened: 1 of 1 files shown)


tfesenko:
@andylowry , I need this change for an integration build. Can you please review it when you the time? Thanks!

andylowry:
@tfesenko Looks good to me

Issue #175, opened 2016-08-08
Optimized, common selection algorithm for all navigation methods


tedepstein:
Outline view, quick outline view and hyperlink navigation are all supported navigation methods in SwagEdit.  There may eventually be others, such as view-to-editor navigation or positional sync with Swagger-UI, or with RepreZen's live Diagram and Documentation views.

We have discussed a couple of optimizations to the navigation algorithm from [outline ](https://github.com/RepreZen/SwagEdit/issues/55) and [quick outline](https://github.com/RepreZen/SwagEdit/commit/9209f72c069073a12318771bea38628db18d6c5c).  But these should be unified to a single, centralized selection behavior.

I would suggest the following rule:
- If the target element constitutes the entire value of a property, select the entire line containing the property name.
- If the target element is not the entire value of a property (e.g. it's a fragment file, a top-level object in a file, or an array element, select the first line of the target element value.

tedepstein:
Mentioned by Saul [here](https://support.reprezen.com/helpdesk/tickets/148).

PR #258, opened 2017-01-11
#175 - Optimized, common selection algorithm for all navigation methods

tfesenko:
swagger_selection

### DIFF
--- a/com.reprezen.swagedit/src/com/reprezen/swagedit/model/AbstractNode.java
+++ b/com.reprezen.swagedit/src/com/reprezen/swagedit/model/AbstractNode.java
@@ -167,6 +167,7 @@ public abstract class AbstractNode {
      * @return position inside the document
      */
     public Position getPosition(IDocument document) {
+        boolean selectEntireElement = false;
         int startLine = getStart().getLine();
         int offset = 0;
         int length = 0;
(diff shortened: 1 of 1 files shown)


tfesenko:

com.reprezen.swagedit/src/com/reprezen/swagedit/model/AbstractNode.java
```diff
@@ -175,7 +176,11 @@ public Position getPosition(IDocument document) {
         int endCol = getEnd().getColumn();
         try {
             offset = document.getLineOffset(startLine);
-            length = (document.getLineOffset(endLine) + endCol) - offset;
+            if (selectEntireElement) {
+                length = (document.getLineOffset(endLine) + endCol) - offset;
+            } else {
+                length = document.getLineOffset(startLine+1) - offset;
```
andylowry, line 182: Coulld this throw BadLocationException if startLine is the last line in the document?

andylowry (commented):
@tfesenko One question for you to consider and handle as needed. Otherwise looks good.

andylowry:
@tfesenko I've fixed the edge case I pointed out. It turns out it's very unlikely. You need a single-line object and no line terminator. Easiest way to test is with an empty method, e.g.:

```
  path /foo:
    get: {}
````
If there's no following line-terminator, then the old code hits BadLocationException and flips back to the top of the editor. With the fix you get the get method highlighted as expected. Obviously not a very important edge case in retrospect, but it's fixed, so please review / test ASAP so we can get a release out for Finnova.

Thanks

PR #259, opened 2017-01-12
#175 - fix tests broken by changing AbstractNode's position

tfesenko:


andylowry (approved):
@tfesenko All changes look reasonable. I'll leave it to the build to decide whether you got all the numbers correct (I expect you did!)

Issue #257, opened 2017-01-11
Support JSON References in contexts not allowed by Swagger 2.0


tedepstein:
One of our users wants to be able to use $ref (JSON reference) properties to re-use centrally defined security schemes and security requirements.

Swagger-Normalizer (our pre-processor) supports this, and inlines the reference correctly, but SwagEdit marks these as errors.  We should allow the user to reduce the severity to a warning, or ignore these errors. 

I'd suggest adding a "Validation" page under the SwagEdit category in Preferences.  Initially it will just have this one option:

Label: JSON Reference property in unsupported context  
Drop-Down list options: Error, Warning, Ignore
Default value: Error

tedepstein:
Resolved in #261 .

PR #261, opened 2017-01-18
#257 Support JSON References in contexts not allowed by Swagger 2.0

tfesenko:
#257 User-configurable severity for unsupported $ref context 
See proposed GUI description inZEN-3180. 
New preferences page:
screen shot 2017-01-18 at 12 44 34 pm

Screencast:
validation_ref

Swagger spec used in the screencast demo:
```yaml
swagger: '2.0'
info:
  version: '2.0.0'
  title:  Some API
  description: Provides an interface for something
basePath: /usage/2
consumes:
  - application/json
produces:
  - application/json
    
paths: 
  /resource:
    get:
 # Security Requirement Object
      security: 
        - $ref: './security.yaml#/security/0' 
      responses:
        '200':
          description: OK

# Security Requirement Object
security: 
  - api_key_query_token: []
  - api_key_header: []
  - $ref: './security.yaml#/security/0' 

# Security Requirements Array
#security: 
#  $ref: './security.yaml#/security'  
 

#  Security Scheme Object
#securityDefinitions: 
#  api_key_query_token:
#    $ref: './security.yaml#/securityDefinitions/api_key'   

# Security Definitions Object
securityDefinitions: 
  $ref: './security.yaml#/securityDefinitions'   
```

Referenced security.yaml:
```yaml
swagger: '2.0'
info:
  version: '2'
  title: Common Security Objects
paths: {}

securityDefinitions:

  api_key:
    type: apiKey
    name: api_key
    in: header

  petstore_auth:
    type: oauth2
    authorizationUrl: http://swagger.io/api/oauth/dialog
    flow: implicit
    scopes:
      write:pets: modify pets in your account
      read:pets: read your pets
  
security:
  - api_key: []
  - petstore_auth: []

```

### DIFF
--- a/com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerEditor.java
+++ b/com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerEditor.java
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package com.reprezen.swagedit.editor;
 
+import static com.reprezen.swagedit.preferences.SwaggerPreferenceConstants.ALL_VALIDATION_PREFS;
+
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


tedepstein:
@tfesenko , please let me know when there's a SwagEdit or API Studio build I can test.  Did you want @andylowry  to do code review?

com.reprezen.swagedit/src/com/reprezen/swagedit/schema/schema.json
```diff
@@ -1158,18 +1158,24 @@
...
+      "oneOf": [
+	    {
+	    "type": "array",
+	     "items": {
+	     "oneOf": [ {
+	          "$ref": "#/definitions/securityRequirement"
+	       }]
+	     }  
+	   }
+	   ],
```
tedepstein, line 1170: Clever!

tedepstein (commented):
Looked through the code.  It took me a few minutes to digest the approach you took here, but it makes sense.  Nice!

tfesenko:
@andylowry , can you please review and QA? 
@andylowry the preference page will certainly look differently on Windows, could you please also post Windows screenshots?

com.reprezen.swagedit/src/com/reprezen/swagedit/schema/SwaggerSchema.java
```diff
@@ -143,6 +147,40 @@ protected void init() {
...
+        if (allow) {
+            definition.add(refNode.deepCopy());
+        } else {
+            int lastIndex = definition.size() - 1;
+            JsonNode lastElement = definition.get(lastIndex);
+            if (refNode.equals(lastElement)) {
+                definition.remove(lastIndex);
+            }
+        }
     }
```
andylowry, line 211: Very cool solution!

com.reprezen.swagedit/src/com/reprezen/swagedit/preferences/SwaggerValidationPreferences.java
```diff
@@ -31,6 +33,7 @@
         implements IWorkbenchPreferencePage, IPropertyChangeListener {
 
     public SwaggerValidationPreferences() {
+        // GRID is needed because we are not attaching the editor fields directly to FieldEditorParent, but to its child
```
andylowry, line 35: This does not make the group work properly in FieldEditorPreferencePage. The problems are much deeper than that. It works properly in certain conditions, but not all. Adding another group or fields with different column requirements will likely lead to unfixable format issues, which is why I'd recommend a comment to this effect here. This breaks the assumption of the FEPP class.

The eclipse documentation on FEPP is misleading on this. It does not point out the risks of using your own layouts, and in fact the FEPP javadoc invites you to override #createLayout if you have special layout needs. This article is closer to correct: http://www.eclipse.org/articles/Article-Field-Editors/field_editors.html. Still, it makes it sound like you can use your own field editors at liberty within your own custom layout. Even that's not strictly true. The bottom line is that FieldEditors all manipulate the containing grid's layout, in ways that cannot be managed properly. So you can use FieldEditors within your own layout, as long as that layout puts every group of "sibling" field editors under the control of the same GridLayout, with nothing else in that GridLayout.

com.reprezen.swagedit/src/com/reprezen/swagedit/schema/SwaggerSchema.java
```diff
@@ -171,13 +173,15 @@ public void allowJsonRefInContext(String jsonReferenceContext, boolean allow) {
         if (definition == null) {
             throw new IllegalArgumentException("Invalid JSON Reference Context: " + jsonReferenceContext);
         }
-        JsonNode refNode = mapper.createObjectNode().put("$ref", "#/definitions/elementRef");
+        int lastIndex = definition.size() - 1;
```
andylowry: I'm not crazy about the assumption that this will always be the last oneOf element. I know that in current code it will be, but perhaps it's an assumption that will be unwittingly violated in some future code, leading to sporadic failures. Might be better to do a List#contains test on the entire oneOf array. Slightly slower, but this only happens when preferences change.

andylowry (approved):

andylowry:
@tfesenko A couple of recommendations in my review, but neither is mandatory.

com.reprezen.swagedit/src/com/reprezen/swagedit/preferences/SwaggerValidationPreferences.java
tedepstein: @tfesenko , @andylowry , BTW, if we were to start this over again, I might have suggested using a checked listbox instead of a group of static listbox controls.  Only because it makes it easier (I think) to add new contexts. 

If you think the edge cases, where the current layout might break, warrant a fix, consider using a checked list as a possible alternative.  But at this point, please choose whichever fix is easiest.  (Or choose no fix at all, if you think we're OK with the current implementation.)

tedepstein:
@andylowry , all of your comments showed up as "outdated."  I had to expand those to see the comments; they didn't appear in the "Files changed" tab.  Any idea why...?

com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerEditor.java
```diff
@@ -191,7 +191,8 @@ public void propertyChange(PropertyChangeEvent event) {
                 }
             }
```
andylowry, line 192: Test - please ignore

andylowry:
@tedepstein wrote:
> all of your comments showed up as "outdated."

A couple of months ago, a new "Review" feature showed up on GitHub. We've been using it, but I've noticed this problem several times. I have a feeling that GitHub's algorithm for deciding what's out-of-date is either buggy or just not very useful.

@tfesenko Maybe we should just go back to using normal comments rather than the Review feature going forward. When adding a comment from the diff screen, there's an "Add Single Comment" button that seems to do it. E.g. my "Test" comment just now. What do you think?

com.reprezen.swagedit/src/com/reprezen/swagedit/preferences/SwaggerValidationPreferences.java
andylowry: In that case it couldn't be a FieldEditorPreference anyway, unless we implemented a ListBoxFieldEditor (or whetever), since there's no existing FieldEditor for that sort of control.

com.reprezen.swagedit/src/com/reprezen/swagedit/preferences/SwaggerValidationPreferences.java
```diff
@@ -42,17 +45,23 @@ public void init(IWorkbench workbench) {
 
     @Override
     protected void createFieldEditors() {
-        Group group = new Group(getFieldEditorParent(), SWT.SHADOW_OUT);
-        GridLayoutFactory.fillDefaults().applyTo(group);
-        GridDataFactory.fillDefaults().grab(true, false).indent(0, 10).applyTo(group);
-        group.setText("Allow JSON references in additional contexts:");
+        // IMPORTANT: FieldEditorPreferencePage does not work very well with complex layouts and nested widgets.
+        // Consider switching to com.modelsolv.reprezen.generators.ui.preferences.GroupedFieldEditorPreferencePage from
+        // the RepreZen repo before modifying it
```
tfesenko, line 49: @andylowry , the comment with suggestions for future modifications is here

com.reprezen.swagedit/src/com/reprezen/swagedit/schema/SwaggerSchema.java
tfesenko: @andylowry , I addressed your suggestion in  74446fa

andylowry (commented):
@tfesenko Looks good.

andylowry:
@tfesenko QA passed. Merging

PR #265, opened 2017-02-23
ZEN-3183 Add scheduling rule to save jobs

andylowry:
This ensures that these jobs will complete before other jobs that depend
on the affected resources are executed.

These changes are required in order for corresponding changes in API Studio to be effective.


andylowry:
@tfesenko This is ready for review.

PR #267, opened 2017-03-10
ZEN-3347 - SwagEdit: Type-ahead does not work for references

tfesenko:
The content-assist scope was changed on typing new characters, while it
should be changed only on Ctrl+space (or its equivalent).


tfesenko:
@andylowry , can you please review this PR?

andylowry (approved):
@tfesenko Code review and QA both pass. Merging

PR #269, opened 2017-03-29
ZEN-3403 - SwagEdit: Content Assist skips suggestions if it finds exactly one candidate

tfesenko:
@andylowry , can you please review and QA?


com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerSourceViewerConfiguration.java
```diff
@@ -57,7 +57,7 @@ public IContentAssistant getContentAssistant(ISourceViewer sourceViewer) {
         ca.setContentAssistProcessor(processor, IDocument.DEFAULT_CONTENT_TYPE);
         ca.setInformationControlCreator(getInformationControlCreator(sourceViewer));
 
-        ca.enableAutoInsert(true);
+        ca.enableAutoInsert(false);
```
andylowry, line 60: I think this looks right, but it's so darned complicated! I'll just have to test it and see... :)

andylowry:
@tfesenko Passes code review and QA. Merging.

Sorry if this means you'll want to do yet another TP change to API Studio. Maybe this one's enough of an edge case that it can be left for the next release.

tedepstein:
We don't need to do another TP update and release for this, it can wait for the next build.

PR #273, opened 2017-04-13
ZEN-3423 Performance optimizations

tfesenko:
## Improvements
1. Discard previous validations on the same model if a new validation is scheduled. I also made the validation job “cancellable” - it means that now cancelling it in the GUI will actually make effect.
2. Model parsing was another memory-consuming operation because Jackson deserialization uses mutant-factories to provide thread-safeness. 
3. Change priority of the `SafeWorkspaceJob` to LONG. We use it for validation and model save, and these are long operations.

## First validation is always slow
Because we need to load JSON Schemas for Swagger and JSON Schema Core, it takes some time. As it’s a one-time thing, it’s out of scope of this PR.

## How to test
You need a big model, a good big model is attached to https://modelsolv.atlassian.net/browse/ZEN-2583 . It also has many validation warnings, it's good test performance of validation.
You may also need to decrease VM memory, as low as `-Xms200m -Xmx200m -XX:PermSize=100m -XX:MaxPermSize=100m` is sufficient.

### DIFF
--- a/com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerDocument.java
+++ b/com.reprezen.swagedit/src/com/reprezen/swagedit/editor/SwaggerDocument.java
@@ -12,6 +12,8 @@ package com.reprezen.swagedit.editor;
 
 import java.io.IOException;
 import java.io.StringReader;
+import java.util.Timer;
+import java.util.TimerTask;
 
 import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.Document;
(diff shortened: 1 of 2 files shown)


tfesenko:

andylowry (approved):
Code review passes

andylowry:
@tfesenko Code looks good.

In QA I was unable to see a noticeable difference between master branch of SwagEdit and this branch. Despite much furious editing and saving of files, I only got the pop-up once, and (alas) that was with the new code. I was running with half the memory you suggested (100Mb heap, 50Mb perm).

I was running with progress view visible, and I rarely saw more than one process active, with either the old or new code. I had all model files in the project you referred to open at all times.

So overall, I think the changes you've made are sensible - and I certainly didn't see any degradation or other problems. I just did not experience an improvement. Perhaps on a less powerful machine things would be different.

What I'd suggest is to go ahead and merge these changes and create a build that uses the new SwagEdit, and then perhaps Ted would be able to test on his least powerful machine, and perhaps verify that performance is noticeably improved.

tedepstein:
OK, merging.

tedepstein:
Ah, OK.

Issue #294, opened 2017-05-12
Warning when an object has a property called "properties".


nkolban:
I have a return JSON schema that looks like:
```
{
   "a": "hello",
   "properties": "world",
   "name": "John",
}
```

The error is on the definition of the "properties" field.

I suspect that the parser is confused with the JSON property named "properties" and the JSON schema keyword "properties".

tedepstein:
Neil, thanks for the report.  Can you please post your Swagger-OpenAPI spec, or an example spec that demonstrates this?

nkolban:
Howdy ... thanks that was careless of me.  Attached you will find the `.yaml` file that I am using.  If you look on line `378` you will see that I have a JSON schema property called "properties".   You will also see an error reported on line `265`.  If I rename the property called `properties` on `378` to something else (eg. `Xproperties`) the error on `265` disappears.

[BPM_REST.yaml.txt](https://github.com/RepreZen/KaiZen-OpenAPI-Editor/files/996855/BPM_REST.yaml.txt)

Should you need anything further, don't hesitate to ask.  I can screen share as needed / desired.

tedepstein:
Thanks, @nkolban .  We'll take a look...

tfesenko:
The fix is merged to master, but not released yet.

tfesenko:
@nkolban , thanks! 
We fixed this issue, the fix is available on Eclipse Marketplace, or you can install an update site from http://products-internal.reprezen.com/swagedit/0.8.0.201705231920

PR #302, opened 2017-05-17
[#294] Warning when an object has a property called "properties".

ghillairet:
See #294


com.reprezen.swagedit.tests/src/com/reprezen/swagedit/validation/ValidatorTest.xtend
```diff
@@ -673,4 +673,34 @@ class ValidatorTest {
...
+		      responses:
+		        '200':
+		          description: OK
+		definitions:
+		  Pets:
+		    type: object
+		    properties:
+		      properties:
+		        type: object
+		        properties:
```
tfesenko, line 695: chicken: chicken:  chicken: 
Thank you for the test!

tfesenko (approved):
Looks good!

Issue #289, opened 2017-05-09
Open API v3: Content Assist for References


tfesenko:
https://modelsolv.atlassian.net/browse/ZEN-3442

In OpenAPI 3, all of these are supported as reusable components, each with its own designated subsection within the OpenAPI Components object:

- [x] Schema Definitions
- [x] Parameters
- [x] Responses
- [x] Path Items
- [ ] Produces/Consumes (does not exist in v3)
- [x] Links - used in responses
- [x] Examples
- [x] Request Bodies
- [x] Headers - used in Link and Response
- [x] Callbacks - used in Operation
- [ ] Security Schemes (see #304)

## Tests
- [x] Parameter,
- [x] Path item,
- [x] Callback,
- [x] Header,
- [x] Example,
- [x] Link,
- [x] Request Bodies
- [ ] Schema Definitions
- [ ] Responses

We already provided support for references to `Schema Definitions` and `Path Items`, but other components need their own `ContextTypes`, see `com.reprezen.swagedit.openapi3.assist.OpenApi3ReferenceProposalProvider`.

ghillairet:
@tfesenko I added support for parameters, responses, requestBodies and links.

You can try with these components

```yaml
components:
  links:
    UserCommitHistory:
      operationId: userCommitHistory
      parameters:
        username: $response.body#/user/username
        page: 0
    
  requestBodies:
    rq:
      description: user to add to the system
      content: 
        'application/json':
          schema:
            $ref: '#/components/schemas/User'
          examples:
            user:
              summary: User Example
              externalValue: 'http://foo.bar/examples/user-example.json'
  responses:
    '200':
      description: Ok
  parameters:
    petId:
      name: petId 
      in: path
      required: true
      description: The id of the pet to retrieve
      schema:
        type: string
```

Parameter reference

```yaml
/pets/{petId}:
  get:
    summary: Info for a specific pet      
    parameters:
      - $ref: "#/components/parameters/petId"
```

Response reference

```yaml
/pets/{petId}:
  get:
    responses:
      200:
        $ref: "#/components/responses/200"
```

Request Body reference

```yaml
post:
  summary: Create
  description: Create a new item
   requestBody: 
     $ref: "#/components/requestBodies/rq"
```

Link reference

```yaml
responses:
  300:
    description: error
    links:
      myLink:
        $ref: "#/components/links/UserCommitHistory"
```

tedepstein:
related issue: #324

PR #311, opened 2017-05-22
#289: Open API v3: Content Assist for References

tfesenko:
Initial implementation of code assist for references.
It also provides tests for code assist for references to:
* parameter,
* path item,
* callback,
* header,
* example
* link

### DIFF
--- a/com.reprezen.swagedit.openapi3/src/com/reprezen/swagedit/openapi3/assist/OpenApi3ReferenceProposalProvider.java
+++ b/com.reprezen.swagedit.openapi3/src/com/reprezen/swagedit/openapi3/assist/OpenApi3ReferenceProposalProvider.java
@@ -18,15 +18,17 @@ public class OpenApi3ReferenceProposalProvider extends JsonReferenceProposalProv
 	public OpenApi3ReferenceProposalProvider() {
 		super(OPEN_API3_CONTEXT_TYPES);
 	}
-
+    
+	protected static final String COMPONENT_NAME_REGEX = "[\\w\\.\\-]+";
+	
 	protected static final String SCHEMA_DEFINITION_REGEX = "^/components/schemas/(\\w+/)+\\$ref|.*schema/(\\w+/)?\\$ref";
 	protected static final String PATH_ITEM_REGEX = "/paths/~1[^/]+/\\$ref";
     protected static final String PARAMETER_REGEX = ".*/parameters/\\d+/\\$ref";
     protected static final String RESPONSE_REGEX = ".*/responses/\\d+/\\$ref";
     protected static final String REQUEST_BODY_REGEX = ".*/requestBody/\\$ref";
     protected static final String LINK_REGEX = ".*/links/\\w+/\\$ref";
-    protected static final String EXAMPLE_REGEX = ".*/examples/[\\w\\.\\-]+/\\$ref";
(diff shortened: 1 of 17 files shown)


com.reprezen.swagedit.openapi3/src/com/reprezen/swagedit/openapi3/assist/OpenApi3ReferenceProposalProvider.java
```diff
@@ -18,15 +18,52 @@
...
+	protected static final String COMPONENT_NAME_REGEX = "[\\w\\.\\-]+";
+	
+    protected static final String SCHEMA_COMPONENT_REGEX = "^/components/schemas/(\\w+/)+";
+    protected static final String INLINE_SCHEMA_REGEX = ".*schema/(\\w+/)*";
+	protected static final String SCHEMA_DEFINITION_REGEX = SCHEMA_COMPONENT_REGEX + "\\$ref" + "|" + INLINE_SCHEMA_REGEX + "\\$ref";
+    protected static final String PATH_ITEM_REGEX = "/paths/~1[^/]+/\\$ref"// in paths object
+            // e.g. "/components/callbacks/myWebhook/$request.body#~1url/$ref"
+            + "|/components/callbacks/" + COMPONENT_NAME_REGEX + "/[^/]+/\\$ref"; // in callbacks
+    protected static final String PARAMETER_REGEX = ".*/parameters/\\d+/\\$ref";
+    protected static final String RESPONSE_REGEX = ".*/responses/\\d+/\\$ref";
```
andylowry, line 31: Status codes in Responses Object can include digits as well as "X" - see https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#patterned-fields-4

com.reprezen.swagedit.openapi3/src/com/reprezen/swagedit/openapi3/assist/OpenApi3ReferenceProposalProvider.java
```diff
@@ -18,15 +18,52 @@
...
 		super(OPEN_API3_CONTEXT_TYPES);
 	}
-
-	protected static final String SCHEMA_DEFINITION_REGEX = "^/components/schemas/(\\w+/)+\\$ref|.*schema/(\\w+/)?\\$ref";
-	protected static final String PATH_ITEM_REGEX = "/paths/~1[^/]+/\\$ref";
-
+    
+	protected static final String COMPONENT_NAME_REGEX = "[\\w\\.\\-]+";
+	
+    protected static final String SCHEMA_COMPONENT_REGEX = "^/components/schemas/(\\w+/)+";
```
andylowry, line 24: I think here and elsewhere when you have `\w` you need to use `COMPONENT_NAME_REGEX`

andylowry (approved):
@tfesenko The hard part of this change is getting all the regexes right. I've pointed out a couple of places where I think they need to be tinkered. It's hard to be confident that all cases are covered, and I'm not sure what the  best way is to gain that confidence. I think it's difficult to reverse-engineer this from the specification, and there may be some ambiguous cases. I don't think the specification was designed with this in mind.

Probably best approach is to do the best we can and address issues as they appear.

In that vein, I'll approve this now, even with the couple of potential corrections I've pointed out, as it should work as-is in the majority of cases. I'd encourage you to open separate issues if you think it's warranted from my comments.

Issue #268, opened 2017-03-27
Content assist typeahead should be case-insensitive


tedepstein:
Pull request #267 makes type-ahead work, but it's case-sensitive.  If you type "pet" when the references are "Pet" and "Pets", the suggestion list disappears, because you've typed something that doesn't have a case-sensitive match to any of the suggestions.

Generally, I think the type-ahead behavior for both keywords and references should be case-insensitive.

tedepstein:
Marked as high-priority, because in combination with #316, content assist is very unforgiving.  Fixing this should be a big usability improvement.

tedepstein:
Tested and verified the fix.  Closing.

PR #318, opened 2017-05-24
#268 - Content assist typeahead should be case-insensitive

tfesenko:
Typeahead on OpenAPI v3 models:
ref_assist1
Similar on Swagger v2:
ref_assist2


tfesenko:

ghillairet (approved):
Looks good

PR #325, opened 2017-05-24
Content Type for v3 should allow unquoted openapi values, e.g. 'openapi: 3.0.0'

tfesenko:
`openapi: 3.0.0` is a perfectly valid value because it is a string. An example of OpenAPI v3 using unquoted version is [authentiq3.yaml](https://gist.githubusercontent.com/MikeRalphson/72e8172bddfa4de7330c9e218b95401f/raw/44600959272b227f6cd12be1f2f604b5f8ac8a7c/authentiq3.yaml)

It was not like this for Swagger v2 because it used a major.minor versioning and `2.0` is a number, not a string.


ghillairet (approved):
Looks correct, it only matches `openapi: '3.0.0` or `openapi: "3.0.0`. it doesn't look for the quote at the end?

tfesenko:
@ghillairet , I addressed your comment + added tests.
Please review again.

ghillairet (approved):
Looks great

PR #326, opened 2017-05-24
Delete unused SwagEdit libs

tfesenko:
I removed com.reprezen.swagedit/lib/ folder from the classpath because they are not needed, they are stored in the core plugin now. 
Maven build passes, I also tested in manually, including RepreZen code generation and live views.


ghillairet (approved):
Looks good.

Issue #321, opened 2017-05-24
Code template for callbacks in operation


tfesenko:
Related to #320 
We should provide a code template for callbacks inside operations:
<img width="417" alt="screen shot 2017-05-24 at 10 14 53 am" src="https://cloud.githubusercontent.com/assets/644582/26407865/45e45380-406a-11e7-923b-33d29ff7b87a.png">

ghillairet:
I updated the templates to use the same callback templates in both operations and components.

See PR #331

![callbacks](https://cloud.githubusercontent.com/assets/148045/26489335/44ef19f8-4207-11e7-89a4-eb02277f6dc5.gif)

PR #330, opened 2017-05-26
[#320] Code assist for callbacks in an operation creates an invalid o…

ghillairet:
…bject

See #320 

Fixed by using a ContentAssistExt to override default behaviour of content assist for callbacks in operations.

PR #331, opened 2017-05-26
[#321] Code template for callbacks in operation

ghillairet:
See #321


tfesenko:
Looks good! 
This PR in combination with PR #330 makes editing operation callbacks so much better!

Issue #313, opened 2017-05-22
OpenAPI v3: code assist for the values of `type`


tfesenko:
Code-assist for the schema type only shows that the value should be a string value:
<img width="240" alt="screen shot 2017-05-22 at 3 23 25 pm" src="https://cloud.githubusercontent.com/assets/644582/26324940/e0d4aaac-3f02-11e7-88d2-1df2d55fe9f6.png">

Code assist for v2 shows a list of types:
<img width="264" alt="screen shot 2017-05-22 at 3 23 39 pm" src="https://cloud.githubusercontent.com/assets/644582/26324961/fd22cd06-3f02-11e7-9c77-23a87cbddbe9.png">

ghillairet:
In v2, the property `type` in schema definition is defined as this

```json
"type": {
  "$ref": "http://json-schema.org/draft-04/schema#/properties/type"
},
```

That reference points to an anyOf definition

```json
"type": {
   "anyOf": [
     { "$ref": "#/definitions/simpleTypes" },
     {
       "type": "array",
       "items": { "$ref": "#/definitions/simpleTypes" },
       "minItems": 1,
       "uniqueItems": true
      }
    ]
}
```

While in v3, the type property is simply defined as a string

```json
"type": {
  "type": "string"
}
```

PR #263, opened 2017-02-13
#262 Use of type: null masks subsequent errors

tfesenko:
## Tests
`NullValueValidationTest` checks for invalid `null` values as well as cases when `null`s are actually valid values in vendor extensions and examples.

## Screenshots
Null definition type has a special message suggesting to use the `"null"` string:
screen shot 2017-02-13 at 1 04 55 pm

Other null value have a standard message:
screen shot 2017-02-13 at 1 03 35 pm

PR #333, opened 2017-05-26
[#313] OpenAPI v3: code assist for the values of `type`

ghillairet:
See #313 where I comment on why it doesn't work.

I solved that by using a ContentAssistExt, this branch uses not yet merged code from #330 

screen shot 2017-05-26 at 16 25 03


com.reprezen.swagedit.openapi3/src/com/reprezen/swagedit/openapi3/assist/ext/SchemaTypeContentAssistExt.java
```diff
@@ -0,0 +1,33 @@
...
+        return type != null && pointer.equals(type.getPointer());
+    }
+
+    @Override
+    public Collection<Proposal> getProposals(TypeDefinition type, AbstractNode node, String prefix) {
+        return Lists.newArrayList( //
+                new Proposal("array", "array", null, "enum"), //
+                new Proposal("boolean", "boolean", null, "enum"), //
+                new Proposal("integer", "integer", null, "enum"), //
+                new Proposal("null", "null", null, "enum"), //
```
tfesenko: I have a déjà vu feeling, that we already discussed null as the type for Swagger v2. Let me look up our old discussions..

tfesenko: As per SUM discussion, we use a `"null"` string (quoted) instead of a `null` value, see `com.reprezen.swagedit.core.assist.JsonProposalProvider.createPropertyProposal(String, TypeDefinition)`
```java
        if (key == "null") {
            key = "\"" + key + "\"";
        }
```
We even added a validation rule for the null type in Swagger v2, see https://github.com/RepreZen/KaiZen-OpenAPI-Editor/pull/263

tfesenko: I added quotes to the replacement value of the null type. 
I also created issue #334 "OpenAPI v3: validation rule for null schema types"

tfesenko (approved):

Issue #339, opened 2017-05-30
Preload JSON Schema for OpenAPI v3 to avoid ProcessingException


tfesenko:
Extracted from #292 
We merged all our customization to gnostic schema, except for [JSON Schema ID: replace a 404 URL with an empty (#) value](https://github.com/RepreZen/gnostic/commit/16c3f2034818aa27620e3d846715b40fecf831ea)

## What's wrong with using the current schema ID in the current code?
Current ID (`http://openapis.org/v3/schema.json`) points to a 404 page
which is not even a JSON.

It causes this exception in Java JSON Schema Validator:
```
Multiple markers at this line
- content at URI "http://openapis.org/v3/schema.json#" is not valid JSON
- 164 changed lines
```

You can also observe the same exception in JSON Schema Validator [online](https://json-schema-validator.herokuapp.com/) by pasting the JSON Schema with `"id": "http://openapis.org/v3/schema.json#"` and an arbitrary v3 model, e.g. `{"info": "", "openapi": "3", "paths": {}}`.

## Where is the problem: is it in our JSON Schema for v3 or in  Java JSON Schema Validator
The answer is neither. The schema ID should be an URL, but it does not have to be a valid URL. JSON Schema Validator handles such cases with `preloadSchema()` method.

## Why do we need it?
This is the last customization to gnostic schema. With this change, we'll be able to use it as is.

PR #340, opened 2017-05-30
#339 - Preload JSON Schema for OpenAPI v3 to avoid ProcessingException

tfesenko:


tfesenko:
How to QA it?
1. First, update the JSON Schema schema ID  to `"id": "http://openapis.org/v3/schema.json#"` without modifying other code. Open an arbitrary v3 spec, e.g. 
```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
    
paths: {}
```
You will see a validation error on the first line:
screen shot 2017-05-30 at 7 14 48 pm

2. Now, switch to the PR branch, open the same spec - it should be OK now.

tfesenko:
@ghillairet , can you please review and QA?

ghillairet (approved):
Looks good, and QA passed.

ghillairet:
Was it because you had to change the id each time the schema is generated? Because # as id is valid.

tfesenko:
@ghillairet wrote:
> Was it because you had to change the id each time the schema is generated? 
 
Yes, exactly. I want to use JSON Schema from gnostic without any modifications.
> Because # as id is valid.

yes, ...and we even use the `#` as `id` for the core schema.

Issue #336, opened 2017-05-30
Reference code assist does not work for single-quoted $ref values


tfesenko:
Open this model:
```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
    
paths:
  /resource:
    get:
      description: description
      responses:
        '200':
          description: Ok
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: ''
  
components:
  schemas:
    Pet:
      properties:
        id:
          type: integer
          format: int64
```
Invoke code assist for the $ref on line 20, nothing will be shown:
<img width="344" alt="screen shot 2017-05-30 at 3 52 53 pm" src="https://cloud.githubusercontent.com/assets/644582/26602053/30debd82-4550-11e7-96ea-f9035df5aa43.png">

Replace the single-quotes with double-quotes. Code assist works now:
<img width="481" alt="screen shot 2017-05-30 at 3 52 58 pm" src="https://cloud.githubusercontent.com/assets/644582/26602076/423afd2a-4550-11e7-85eb-98ec1c0ab6d3.png">

## Why is it important?
Many of our code templates, e.g. Object Resource templates, use single-quotes for $refs.

PR #341, opened 2017-05-31
[#336] Reference code assist does not work for single-quoted $ref values

ghillairet:
See #336 

quotes


tfesenko:
Nice!
Passed code review and QA.

Issue #323, opened 2017-05-24
OAS3: Validation of $ref property values should show error on unexpected object type


tedepstein:
Validation seems happy as long as it can resolve each `$ref` property to a valid JSON or YAML object.  It should validate that the referenced object is of the expected type, meaning that it conforms to the JSON schema that can be used in place of the $ref.

tfesenko:
Often times (always?), JSON Schema provides needed information for references as references are defined inside an `xyzOrReference` element:
```json
"schemas": {
          "$ref": "#/definitions/schemasOrReferences"
        }
```

ghillairet:
See PR #342 

Error is now shown when reference is not of expected type. (I haven't tested yet on all possible kind of references). Also doesn't work yet for references that are not in a $ref property.

![validation](https://cloud.githubusercontent.com/assets/148045/26635747/0ffcab58-461b-11e7-81e4-8c646c0557a6.gif)

tfesenko:
As discussed at today's SUM, this issue only covers the `$ref` references, it does NOT cover securityScheme, operationRef and operationId. The focus is on OpenAPI v3, and support for v2 is optional, we just should not break it :) 

Another topic that we discussed at today's SUM is more tests. We need to check for
* No false negatives: a valid model should not be marked as invalid (see my comment about schema arrays)
* No false positives: invalid models should not be marked as valid.

How can we test it all? 
## $ref usages
`$ref` references can substitute the following elements:
Components which are usually described with the "xyzOrReference":
 * Schema Definitions
 * Parameters
 * Responses
 * Links
 * Examples
 * Request Bodies
 * Headers
 * Callbacks
 * Responses

PLUS

* Path Items, where the `$ref` is described along with the other properties:
 ```json
        "$ref": {
          "type": "string"
        },
        "summary": {
          "type": "string"
        },
```

OpenAPI v3 also allows references to `example` (singular), which is a freeform example. It's different from the `examples` object which also provides the name and/or description.

Each of these 10 elements can be referenced in different contexts. I created OpenAPI v3 specs demonstrating all major use cases for reference in com.reprezen.swagedit.openapi3.tests/resources/code-assist/references .  Schema ref example has too many new cases, e.g. for `oneOf`, `not`, or `anyOf` and is under development. I will also create an example for responses references. 

### False negatives
[...]

Issue #308, opened 2017-05-19
OpenAPI v3: Validation for the value of Link#operationId


tfesenko:
Related to #307 

[Link#operationId](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#link-object) is
> the name of an existing, resolvable OAS operation, as defined with a unique operationId. This field is mutually exclusive with the operationRef field. Relative operationRef values MAY be used to locate an existing Operation Object in the OAS.

Validation should verify that an operation with operationIds exists.

To clarify: can we reference operationIds from external models? 
Answer: No, only local elements.
>  In the case of an operationId, it MUST be unique and resolved in the scope of the OAS document

ghillairet:
@tfesenko I think this one can be closed.

Behaviour seen on master

![sc](https://user-images.githubusercontent.com/148045/30963329-577dbc5c-a44d-11e7-9acf-8dd4195db07b.gif)

tfesenko:
@ghillairet , thanks!
I can see a validation error too. The current error message is `Invalid object reference, the referenced object is not of expected type`, it can be improved and provide information that an operation with this operationId does not exist at all. 

<img width="515" alt="screen shot 2017-09-28 at 9 05 59 am" src="https://user-images.githubusercontent.com/644582/30968143-54085f6c-a42c-11e7-8429-b0f42db9f8ae.png">

Issue #306, opened 2017-05-19
OpenAPI v3: Validation for Link#operationRef values


tfesenko:
Related to #305 

[Link#operationRef](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#link-object) is
> a relative or absolute reference to an OAS operation. This field is mutually exclusive with the operationId field, and must point to the fragment of a valid OAS definition.

Validate if referenced value exists and it's an operation.

ghillairet:
@tfesenko Validation works when operationRef is a local reference but fails when it's an external one (absolute ref).

ghillairet:
I would do it once #358 is merged since it is likely to reuse code from it.

ghillairet:
See PR #414

PR #342, opened 2017-05-31
[#323] Validation of $ref property values should show error on unexpe…

ghillairet:
…cted object type

See #323 

Probably needs more tests.


tfesenko:
@ghillairet , 
This approach does not work for arrays:
screen shot 2017-05-31 at 10 27 14 am

I used this spec:
```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
    
paths: {}
                
components:
  schemas:
  
    MyType2:
      type: array
      items:
        $ref: "#/components/schemas/MyType1"
    MyType1:
      required:
        - property
      properties:
        property:
          type: integer
          format: int64     
```

com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/validation/Validator.java
```diff
@@ -266,6 +269,31 @@ protected void checkMissingRequiredProperties(Set<SwaggerError> errors, Abstract
...
+            if (type instanceof ReferenceTypeDefinition) {
+                type = ((ReferenceTypeDefinition) type).resolve();
+            }
+
+            AbstractNode nodeValue = node.get(JsonReference.PROPERTY);
+            AbstractNode valueNode = model.find((String) nodeValue.asValue().getValue());
+
+            if (valueNode != null && type.getPointer() != null && valueNode.getType() != null
+                    && !type.getPointer().equals(valueNode.getType().getPointer())) {
+                errors.add(error(nodeValue, IMarker.SEVERITY_ERROR, Messages.error_invalid_reference_type));
```
tfesenko: It would be **really** helpful to provide what type is actually expected. It looks like `valueNode.getType().getPointer()` can provide this info

ghillairet:
Ok I'll fix the array and add the type in the error.

com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/validation/Validator.java
```diff
@@ -288,6 +290,26 @@ protected void checkMissingRequiredProperties(Set<SwaggerError> errors, Abstract
...
+     * @param errors
+     * @param node
+     */
+    protected void checkReferenceType(Set<SwaggerError> errors, AbstractNode node) {
+        if (JsonReference.isReference(node)) {
+            Model model = node.getModel();
+            TypeDefinition type = node.getType();
+
+            AbstractNode nodeValue = node.get(JsonReference.PROPERTY);
+            AbstractNode valueNode = model.find((String) nodeValue.asValue().getValue());
```
tfesenko, line 309: It only works for local references. It throws and exception for external references:
* From the same workspace (local external reference via relative path)
* Or, remove (accessed via HTTP/HTTPS/any other protocol)

Because it throws an exception which is never processed, the exception is propagated to the top level `com.reprezen.swagedit.core.validation.Validator.validate(JsonDocument, URI)` and it means that validation errors are not processed at all.

tfesenko:
@ghillairet , this code works well with references to the elements from the same file.

However, I found a severe problem with JSON pointers which are not starting with a "/". So, if such pointer is present in a spec, validation does NOT work at all. You can type anything and it will be accepted by the validator, see my comment https://github.com/RepreZen/KaiZen-OpenAPI-Editor/pull/342#discussion_r119873194
Examples of such pointers:
* VALID REF: References to the elements from the same workspace, e.g. ` $ref: "uber_schemas.yaml#/components/schemas/PriceEstimate"`. https://modelsolv.atlassian.net/secure/attachment/33000/Multi-file%20Uber%20%28OAS3%29.zip has good examples. 
* VALID REF: References to remote resources, e.g. `$ref: "https://raw.githubusercontent.com/RepreZen/KaiZen-OpenAPI-Editor/master/com.reprezen.swagedit.openapi3.tests/resources/spec_examples/v3.0/uber.yaml#/components/schemas/PriceEstimate"`
* INVALID REF: just invalid pointer, like in my JUnit test: `$ref: "INVALID"`. Yes, it is invalid, but it should show a warning, instead of breaking validation.

All these cases are caused by the same line of code. But we should test them (preferably with automated tests) separately after fixing it. 
Exception:
```java
java.lang.IllegalArgumentException: Invalid input: JSON Pointer expression must start with '/': "./components/uber_path_products.yaml#/paths/~1products2"
	at com.fasterxml.jackson.core.JsonPointer.compile(JsonPointer.java:122)
	at com.fasterxml.jackson.core.JsonPointer.valueOf(JsonPointer.java:131)
	at com.reprezen.swagedit.core.model.Model.find(Model.java:222)
	at com.reprezen.swagedit.core.validation.Validator.checkReferenceType(Validator.java:305)
	at com.reprezen.swagedit.core.validation.Validator.validateModel(Validator.java:175)
	at com.reprezen.swagedit.core.validation.Validator.validate(Validator.java:118)
	at com.reprezen.swagedit.core.validation.Validator.validate(Validator.java:100)
```

tfesenko:
I added more tests for invalid refs, so the old tests + new tests cover these variations of invalid refs:
- [x] not a pointer
- [x] points to a top level element which is not a JSON
- [x] pointer to a nested element inside a non-JSON file
- [x] points to a JSON of invalid type (covered by initial set of tests)

The tests check that these scenarios don't break the validation

ghillairet:
@tfesenko I updated this PR to make your tests pass. The validation of remote refs as also been fixed, and support for links operationId, operationRef and securitySchemes (see #308 , #306 and #347)

ghillairet:
There is still an issue with validation of refs inside the same workspace/project, I'm trying to fix that now.

com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/validation/Validator.java
```diff
@@ -169,13 +170,18 @@ public Validator(JsonReferenceValidator referenceValidator, Map<String, JsonNode
...
             for (AbstractNode node : model.allNodes()) {
-                checkArrayTypeDefinition(errors, node);
-                checkObjectTypeDefinition(errors, node);
+                executeModelValidation(model, node, errors);
             }
         }
         return errors;
     }
 
+    protected void executeModelValidation(Model model, AbstractNode node, Set<SwaggerError> errors) {
```
tfesenko, line 179: I suggest `validateNode` as method name

com.reprezen.swagedit.openapi3/src/com/reprezen/swagedit/openapi3/assist/contexts/OperationIdContextType.java
```diff
@@ -27,12 +27,7 @@ public OperationIdContextType(CompositeSchema schema, String regex) {
     @Override
     public Collection<Proposal> collectProposals(Model model, IPath path) {
         final Collection<Proposal> results = Lists.newArrayList();
-        final List<AbstractNode> nodes = Lists.newArrayList();
-        for (AbstractNode node : model.allNodes()) {
-            if (node.getType() != null && operationPointer.equals(node.getType().getPointer())) {
-                nodes.add(node);
-            }
-        }
+        final List<AbstractNode> nodes = model.findByType(operationPointer);
```
tfesenko, line 30: Nice

com.reprezen.swagedit.tests/src/com/reprezen/swagedit/validation/ValidatorTest.xtend
```diff
@@ -704,4 +705,110 @@ class ValidatorTest {
...
+			    type: array
+			    items:
+			      $ref: "#/definitions/Bar"
+			  Bar:
+			    type: object
+		'''
+
+		document.set(content)
+		val errors = validator.validate(document, null as URI)		
+		assertEquals(0, errors.size())
```
tfesenko, line 784: :) thanks

tfesenko (approved):
Passed code review and QA

Issue #343, opened 2017-05-31
Tests: SchemaValidationTest can only be run as JUnit Plugin Test, make it a regular JUnit test


tfesenko:
Extracted from PR #329 
Tests: SchemaValidationTest can only be run as JUnit Plugin Test. Eliminate dependency on OSGi, make it a regular JUnit test.

tfesenko:
Fixed. Thanks, @ghillairet for pointing me to `ErrorProcessorTest`

PR #351, opened 2017-06-06
#343 - SchemaValidationTest can only be run as JUnit Plugin Test, make it a regular JUnit test

tfesenko:


ghillairet (approved):
Looks good

PR #355, opened 2017-06-09
#289 - calculate content assist context type for $ref values based on JSON schema, not regex

tfesenko:
This PR introduces a new `ComponentContextType` which uses JSON schema to retrieve context type for $ref values. 
* It eliminates our public static final regexes
* It also fixes code assist context for schema, as the regex-based version was too relaxed and was using schema contexts for schema examples

Issue #290, opened 2017-05-09
OpenAPI v3: Content Assist Code Templates


tfesenko:
https://modelsolv.atlassian.net/browse/ZEN-3443

We expect to create a new set of default code templates for OpenAPI v3. The code templates are provided based on the code template contexts, so we will need to 
1. properly calculate contexts and 
2. create new contexts for the new elements.

We will also need a preferences page for code templates (see https://modelsolv.atlassian.net/browse/ZEN-3448)

See the `com.reprezen.swagedit.templates` as an example how it's implemented for Swagger v2

tedepstein:
@tfesenko , AFAIK, this is essentially complete; we should just have separate open issues for any contexts that need templates to be added, improved, or that are not showing their templates as expected.

PR #357, opened 2017-06-09
#290: Centralize registration of code template contexts

tfesenko:
Based on changes from PR #355

Prior to this change, we had to register a new code template context in two places:

* In plugin.xml, which was parsed by ContributionContextTypeRegistry
* OpenApi3ContextType#allContextTypes

The old ContributionContextTypeRegistry took a context name as an argument, looked up in plugin.xml items, and registered what the found item.
The new ContextTypeRegistry does not need plugin.xml items as it takes a TemplateContextType directly


ghillairet (approved):
Looks good.

Issue #353, opened 2017-06-08
Validation of external $ref property values should show error on unexpected object type


tfesenko:
Extracted from #323 Validation of $ref property values should show error on unexpected object type

It covers validation of property values which are defined in an external file, but which resides in the same workspace/project, e.g. `$ref: "./uber_schemas.yaml#/components/schemas/Product"`

## [Local] Validation checks that the referenced element is actually what we expect: good for correctness, bad for performance
We check the local references by getting their actual JSON schema type and comparing it to the expected value. It's done by loading the referenced definition and matching it against the JSON schema. It's OK for local references because all elements are already loaded and schema-matched, but it can *really* slow down editing experience as validation is live and loading/schema-matching is executed on every change.

## Solution: "Batch Validate" action
We can add a "Validate References" action. It's OK for it to be slow  - it will now influence editing experience as the users will be invoking it explicitly, and it will still allow us to validate references.

## Complimentary step: Check the JSON pointer first
We know that the referenced elements **must** live at a certain location, e.g. schema definitions must have a `<optional file url>#/components/schemas/{componentName}` pointer, headers must be in `<...>#/components/headers/{componentName}`, etc. 
So, while not all invalid referenced elements have invalid (wrong format) pointers, but all elements with invalid (wrong format) JSON Pointer are invalid values for references.
We can run validation in two steps:
1. Check the JSON Pointer. It will not guarantee that the referenced element actually has the needed fast, or even that the file is a valid OAS 3 file, but it's fast. 
2. For those elements whose JSON pointers are fine, check their actual type

ghillairet:
See PR #358 

I moved the code that does validation of references into the reference validator that was previously only checking that the URI or pointer actually points to something. By adding a few methods I am able to load the referenced document and parse it as a Model. From that I can check the type of the referenced object.

Only missing in this PR is a few tests that I'll add next week.

Also worth to check that this way of validating references does not impact too much the performance of the editor, as multiple documents have to be parsed during validation.

tedepstein:
I ran into this when testing multi-file BeamUp, which has an invalid reference to a parameter object, where reference to a callback object would be expected.   Looks like implementation of the fix is already underway in #358. 

I have a couple of comments:

1. Requiring an explicit, specialized user action to validate references seems like a pretty significant usability compromise.  Could we have an option, enabled by default, to validate references on save? 
[...]

Issue #345, opened 2017-06-02
OpenAPI v3: Validation of *remote* $ref property values


tfesenko:
OpenAPI v3: Validation of *remote* $ref property values should show error on unexpected object type, but should allow references to the expected type.

Extracted from #323 which covers validation for local $refs and $refs using relative paths.
This issue is for remote $refs, e.g. accessed with HTTP/HTTPS: `$ref: "https://raw.githubusercontent.com/RepreZen/KaiZen-OpenAPI-Editor/master/com.reprezen.swagedit.openapi3.tests/resources/spec_examples/v3.0/uber.yaml#/components/schemas/PriceEstimate"`

See https://github.com/RepreZen/KaiZen-OpenAPI-Editor/pull/342#issuecomment-305810054

tedepstein:
It seems that the editor will always warn on an remote reference, or a certain type of external reference, even when valid.  

"Remote reference" means a reference to a document that is not in the current workspace. This problem was observed specifically on an http reference.  We don't know whether the problem applies to all remote references, or just http, or just https, maybe.

Here's a gist, with a valid OpenAPI 2.0 (Swagger) spec containing schemas:
https://gist.github.com/tedepstein/0e5f5982314f0648377df8782d7b7ccf

Here's an OpenAPI 2.0 spec with a reference to a definition in that gist (using the 'raw' URL for the gist):

```
swagger: "2.0"
info:
  description: Tax Blaster
  version: 1.0.0
  title: TaxBlaster
host: taxblaster.com
basePath: /api
schemes:
- http

paths:

  /taxFilings/{id}:

    get:
      tags:
      - TaxFilingObject
      description: Retrieves a tax filing having the specified id.
      operationId: getTaxFiling
      consumes:
      - application/json
      produces:
      - application/json
      parameters:
      - name: id
        in: path
        description: ID of the requested TaxFiling
        required: true
        type: string
      responses:
        200:
          description: Successful response, with a representation of the Tax Filing.
          schema:
            # Reference to a Schema Object described in 'definitions' section
            $ref: "https://gist.githubusercontent.com/tedepstein/0e5f5982314f0648377df8782d7b7ccf/raw/96826757aa77ddd6816ff7c3e5ca6903c0f66f2b/schemas.yaml#/definitions/PriceEstimate"
        404:
          description: The requested tax filing was not found.
```

The reference is flagged with a warning:

```
[...]

PR #358, opened 2017-06-09
[#353] Validation of external $ref property values should show error …

ghillairet:
…on unexpected object type

Validation of reference types has been moved to ReferenceValidator that previously only validate correctness of references pointer/URIs.
It now checks the type of the reference if this one is valid, this allow to reuse the code for loading the JSON document which is referenced by it.
This validation could be slow because it will parse the referenced documents into Models.

See #353


ghillairet:
@tfesenko Can you please review. I updated this PR to include validation of types for external references. See #345 #353

com.reprezen.swagedit.openapi3.tests/src/com/reprezen/swagedit/openapi3/validation/ReferenceValidatorTest.xtend
```diff
@@ -0,0 +1,393 @@
...
+			      in: query
+			      type: string
+		'''
+
+		document.set(content)
+		val baseURI = new URI(null, null, null)
+		val resolvedURI = new URI(null, null, "/components/parameters/Valid")
+		val errors = validator(#{resolvedURI -> document.asJson}).validate(baseURI, document)
+
+		assertEquals(0, errors.size())
```
tfesenko, line 55: Thanks!

com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/json/references/JsonDocumentManager.java
```diff
@@ -91,11 +91,6 @@ public JsonNode getDocument(URL url) {
     }
 
     public JsonNode getDocument(URI uri) {
-        final IFile file = getFile(uri);
-        if (file == null || !file.exists()) {
-            return null;
-        }
-
         try {
             return getDocument(uri.toURL());
```
tfesenko, line 71: We are passing an entire URL to `JsonDocumentManager.getDocument(URL)` which includes both the `file` and `ref` parts. `JsonDocumentManager.getDocument(URL)` has a table whose keys are `URL`s in a form we are passing the here. As a result, for two different refs from the **same** external model will load the model twice. For example, for the `beamup_path_estimates_price.yaml` from the Multi-file Beamup (v3) example the `documents` map will contain a different JsonNode model for each parameter:
```
{<...>/Multi-file%20Beamup%20(v3)/models/components/beamup_parameters.yaml#/components/parameters/start_longitude
    ={"openapi":"3.0.0", <...>}, 

<...>/Multi-file%20Beamup%20(v3)/models/components/beamup_parameters.yaml#/components/parameters/end_longitude
   ={"openapi":"3.0.0",<...>}
}
```
But it should be a single entry for `<...>/Multi-file%20Beamup%20(v3)/models/components/beamup_parameters.yaml`! 
We should fix `JsonDocumentManager.getDocument(URL)`.

com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/json/references/JsonReference.java
```diff
@@ -223,4 +216,19 @@ public static JsonPointer getPointer(ObjectNode node) {
         }
     }
 
+    /**
```
tfesenko, line 220: Javadoc for `JsonReference.isMissing()` should be corrected:
```java
    /**
     * Returns true if the reference points to an existing JSON document and the pointer to an existing node inside that
     * document.
     * 
     * @param baseURI
     * @return true if the reference can be resolved.
     */
    public boolean isMissing(JsonDocument document, URI baseURI) {
        if (isInvalid()) {
            return false;
        }

        JsonNode resolved = resolve(document, baseURI);
        return resolved == null || resolved.isMissingNode();
    }
```
It returns true if the reference can**not** be resolved and the pointer points to an **inexistent** element.

tfesenko (changes requested):
I left several comments including a comment about the `documents` cache.
Also, we have several test failures on this branch:
```
Failed tests: 
  CodeTemplateTextTest.test_code_template_text:148 JSON Schema validation error: 

Tests in error: 
  ValidatorTest.testValidationShouldFail_ForInvalidPointers:314 » IllegalArgument
  ValidatorTest.testValidationShouldFail_refNotJson:368 » Scanner null; mapping ...
```

ghillairet:
@tfesenko The PR has been updated.

tfesenko:
@ghillairet , I am still getting two test failures on this branch (after merge master in to the task branch, the master is green):
```
Failed tests: 
  ValidatorTest.testValidationShouldFail_ForInvalidPointers:317 expected:<1> but was:<2>
  ValidatorTest.testValidationShouldFail_refNotJson:371 expected:<1> but was:<2>

```

tfesenko:
Passed code review and QA.
All OK except for test failures.

ghillairet:
Tests have been fixed.

com.reprezen.swagedit.openapi3.tests/src/com/reprezen/swagedit/openapi3/validation/ValidatorTest.xtend
```diff
@@ -164,8 +164,8 @@ class ValidatorTest {
 			paths: {}
 			components: 
 			  schemas:
-			  
 			    MyType1:
+			      type: object
```
tfesenko, line 168: :smile:

tfesenko:
Looks good. Thanks!!

PR #360, opened 2017-06-29
#290 Use JSON Schema to calculate code template contexts + test framework for code templates

tfesenko:
## Code template contexts driven by JSON Schema
We already did it for the references and, as a natural continuation, I am applying the same approach to Code template contexts. It's implemented in `SchemaBasedTemplateContextType` which defines code template contexts for all components and component maps. Root, contact, path, and path item contexts are still using the old regex-based approach and will be migrated to JSON Schema later.

# Tests
We use the same specification by example approach as for references. I added OAS3 examples to `resources/code-assist/code-templates/`. A context is defined using comments in a `#KZOE-template name="callback in components/callback", value="callback"` format:
```yaml
     MyWebhooWithRef:
      #KZOE-template name="callback in components/callback", value="callback"
       $ref: "#/components/callbacks/myWebhook"		       $ref: "#/components/callbacks/myWebhook"
     myWebhook:		     myWebhook:
```
## Where? Right context at a right place
The tests verify that at the marked location we are getting a code template context as defined in the `value` property of our spec comment.

## What? Existing code templates create a valid model
The tests also collect **all existing code templates** for this context and check that after applying a template the OAS spec is still valid (against the JSON Schema)

## Screenshot
This is what it looks like from Eclipse
screen shot 2017-06-29 at 10 02 53 am
and
screen shot 2017-06-29 at 10 10 44 am

## Coverage
screen shot 2017-06-29 at 9 48 30 am


tfesenko:

tedepstein:
Nice, @tfesenko .  I appreciate the coverage map.

ghillairet (approved):
Looks good

Issue #293, opened 2017-05-12
Eclipse Neon 3 and Reprezen OpenAPI Editor Plugin classloader issue


philwhiles:
Am trying(!) to try out OpenAPI Editor plugin from the Eclipse Marketplace in my Eclipse Neon 3 install.
Dragged to install.

On trying to open a yaml file I get an error and yaml editor fails to open:

java.lang.LinkageError: loader constraint violation: when resolving field "mapper" the class loader (instance of org/eclipse/osgi/internal/loader/EquinoxClassLoader) of the referring class, com/reprezen/swagedit/core/schema/CompositeSchema, and the class loader (instance of org/eclipse/osgi/internal/loader/EquinoxClassLoader) for the field's resolved type, com/fasterxml/jackson/databind/ObjectMapper, have different Class objects for that type
	at com.reprezen.swagedit.schema.SwaggerSchema.<init>(SwaggerSchema.java:23)
	at com.reprezen.swagedit.Activator.getSchema(Activator.java:150)
	at com.reprezen.swagedit.editor.SwaggerDocument.<init>(SwaggerDocument.java:24)
	at com.reprezen.swagedit.editor.SwaggerDocumentProvider.createEmptyDocument(SwaggerDocumentProvider.java:26)
	at org.eclipse.ui.editors.text.StorageDocumentProvider.createDocument(StorageDocumentProvider.java:226)
	at com.reprezen.swagedit.core.editor.JsonDocumentProvider.createDocument(JsonDocumentProvider.java:35)
	at org.eclipse.ui.editors.text.FileDocumentProvider.createElementInfo(FileDocumentProvider.java:721)
	at org.eclipse.ui.texteditor.AbstractDocumentProvider.connect(AbstractDocumentProvider.java:392)
	at org.eclipse.ui.texteditor.AbstractTextEditor.doSetInput(AbstractTextEditor.java:4178)
	at org.eclipse.ui.texteditor.StatusTextEditor.doSetInput(StatusTextEditor.java:229)
	at org.eclipse.ui.texteditor.AbstractDecoratedTextEditor.doSetInput(AbstractDecoratedTextEditor.java:1466)
	at org.eclipse.ui.editors.text.TextEditor.doSetInput(TextEditor.java:150)
	at com.reprezen.swagedit.core.editor.JsonEditor.doSetInput(JsonEditor.java:161)
[...]

PR #367, opened 2017-08-02
#293 Fix classloader issue

tfesenko:
This PR fixes classloader issue described in #293 by using Orbit Jackson and eliminating import-package in MANIFEST.MF
We were previously using Jackson from the RepreZen bundles plugin. I removed it from there and added Tycho-friendly Jackson from Eclipse Orbit 

# How to QA
A followed the steps from [here](https://github.com/RepreZen/KaiZen-OpenAPI-Editor/issues/293#issuecomment-312732132 )
1. Install Eclipse Oxygen. I used a brand-new oomph p2 pool just in case.
2. Install Eclipse Docker Tooling from Eclipse Marketplace
3. Install Spring IDE plugins (STS) from Eclipse Marketplace

To check that the reported classpath problem is present in old builds:
4. Install KaiZen OpenAPI Editor from this update site: http://products-internal.reprezen.com/swagedit/0.8.0.201707202032. This is the latest update site which uses Jackson from RepreZen bundles. 
Open a Swagger file - the editor will not be opened, Eclipse will show an error dialog.

To check that the new build fixes it:
5. Get the same Eclipse+Docker Tools + STS plugins configuration by uninstalling KaiZen editor.
6. Install KaiZen Editor using this update site: http://products-internal.reprezen.com/swagedit/0.8.0.201708021857/ . It's built on the PR.
Open a Swager or OAS3 file - the editor is opened and functioning correctly.

# Overview
## Target Platform 
I completely removed RepreZen Bundles plugin from the target platform.
* I added Jackson from http://download.eclipse.org/tools/orbit/downloads/drops/R20170516192513/ to the update site.
Note: there are newer versions of Orbit libs available, we will migrate to them later.
* SnakeYAML is also a required and shipped by Orbit. A funny thing is that it (SnakeYAML) was listed on the Orbit site, but was not visible in the target platform editor. The reason? They used a wrong label:
screen shot 2017-08-02 at 5 18 39 pm
I found it among other `bundleName`s by version. 

* Because Swagger2 model libs were provided by RepreZeb bundles, I re-added them to the [cr].swagedit/lib.

##  YamlErrorProcessor suffering from non-visible JacksonYAMLParseException and MarkedYAMLException
`JacksonYAMLParseException` and `MarkedYAMLException` are not Tycho-exposed in Orbit. So, had to use a hack to restore proper errors on the wrong indentation. The automated tests pass, I also manually tested it. But any new input is welcome

## `YAMLMapper` cannot be instantiated in Tycho
While `YAMLMapper` is visible, using it throws an exception at runtime. The reason is that Jackson Dataformat library does not declare an OSGi dependency on Jackson Databind which is needed to initialize YAMLMapper (it extends ObjectMapper from Jackson Databind).
Replacing `new YAMLMapper()` with `new ObjectMapper(new YAMLFactory())` solves it.


ghillairet (approved):
Passed QA, works as expected.

PR #368, opened 2017-08-03
Fix type-ahead for OAS3 keywords

tfesenko:
Null `model` was causing an NPE which was propagated to `JsonContentAssistProcessor.getContextType(ITextViewer, IRegion)`. As a result, type-ahead was not working for OAS3 keywords.


ghillairet (approved):
Looks good.

Issue #370, opened 2017-08-09
Missing validation and content assist on `in` property of parameter object


tedepstein:
Thee `in` property should accept only certain values, including `header`, `query`, etc.  Content assist should offer these values, and validation should enure that `in` is set to one of the allowed values. 

In this example, the value "headerzzz" should be disallowed:

```
openapi: 3.0.0
info: 
  title: Callback Example
  version: 1.0.0
paths:
  /streams:
    post:
      description: subscribes a client to receive out-of-band data
      parameters:
        - name: callbackUrl
          in: headerzzz
          required: true
          description: |
            the location where data will be sent.  Must be network accessible
            by the source server
          schema:
            type: string
            format: uri
            example: https://tonys-server.com
      responses:
        '201':
          description: subscription successfully created
```

ghillairet:
Fixed in #376 

![param](https://user-images.githubusercontent.com/148045/29352764-3cbc71b2-8268-11e7-96b7-9462f6ad4b2d.gif)

PR #376, opened 2017-08-16
[#370] Missing validation and content assist on `in` property of para…

ghillairet:
…meter object

### DIFF
--- a/com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/validation/messages.properties
+++ b/com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/validation/messages.properties
@@ -35,6 +35,7 @@ error_missing_properties = Invalid type definition, object has a required field 
 warning_simple_reference = Simplified reference syntax is deprecated. The reference should be a valid JSON pointer.
 error_array_missing_items= Invalid array definition, items type should be present
 error_invalid_reference_type = Invalid object reference, the referenced object is not of expected type
+error_invalid_parameter_location = Invalid parameter location value, possible values are \"query\", \"header\", \"path\" or \"cookie\".
 content_assist_proposal_local = Press '%s' to show %s in the current file.
 content_assist_proposal_project = Press '%s' to show all %s in the project.
 content_assist_proposal_workspace = Press '%s' to show all %s in the workspace.
--- a/com.reprezen.swagedit.openapi3.tests/src/com/reprezen/swagedit/openapi3/validation/ValidatorTest.xtend
+++ b/com.reprezen.swagedit.openapi3.tests/src/com/reprezen/swagedit/openapi3/validation/ValidatorTest.xtend
@@ -26,7 +26,7 @@ class ValidatorTest {
 	val factory = new JsonReferenceFactory() {
 		override create(AbstractNode node) {
 			val reference = super.create(node)
-			if (reference != null) {
+			if (reference !== null) {
 				reference.documentManager = docManager
 			}
 			reference
@@ -34,7 +34,7 @@ class ValidatorTest {
 
 		override createSimpleReference(URI baseURI, AbstractNode valueNode) {
 			val reference = super.createSimpleReference(baseURI, valueNode)
-			if (reference != null) {
+			if (reference !== null) {
 				reference.documentManager = docManager
 			}
 			reference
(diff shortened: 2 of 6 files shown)


com.reprezen.swagedit.openapi3/src/com/reprezen/swagedit/openapi3/validation/OpenApi3Validator.java
```diff
@@ -86,4 +88,23 @@ protected void validateOperationIdReferences(Model model, AbstractNode node, Set
         }
     }
 
+    protected void validateParamters(Model model, AbstractNode node, Set<SwaggerError> errors) {
```
tfesenko: A typo here - `validateParam*e*ters `

com.reprezen.swagedit.openapi3.tests/src/com/reprezen/swagedit/openapi3/validation/ValidatorTest.xtend
```diff
@@ -409,4 +409,62 @@ class ValidatorTest {
...
+			            type: string
+			      responses:
+			        '200':
+			          description: Ok
+		'''
+
+		document.set(content)
+		val errors = validator.validate(document, null as URI)
+
+		assertEquals(0, errors.size())
```
tfesenko, line 468: Thanks for the tests!

com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/validation/messages.properties
```diff
@@ -35,6 +35,7 @@ error_missing_properties = Invalid type definition, object has a required field
 warning_simple_reference = Simplified reference syntax is deprecated. The reference should be a valid JSON pointer.
 error_array_missing_items= Invalid array definition, items type should be present
 error_invalid_reference_type = Invalid object reference, the referenced object is not of expected type
+error_invalid_parameter_location = Invalid parameter location value, possible values are \"query\", \"header\", \"path\" or \"cookie\".
```
tfesenko, line 38: I just noticed that `OpenApi3Validator` from `com.reprezen.swagedit.openapi3` uses `Messages` from `com.reprezen.swagedit.core`. 
It's OK for now, but we should fix it.

tfesenko:
Passed QA and code review. 
@tedepstein , current error message is `Invalid parameter location value, possible values are "query", "header", "path" or "cookie".`. Please let us know if we need to rephrase it or if we need to add a comma before the `or` - according to [this article](https://english.stackexchange.com/a/30519) a comma here is optional, but many people preper to add it.

com.reprezen.swagedit.openapi3/src/com/reprezen/swagedit/openapi3/validation/OpenApi3Validator.java
```diff
@@ -86,4 +88,23 @@ protected void validateOperationIdReferences(Model model, AbstractNode node, Set
...
+        final JsonPointer pointer = JsonPointer.compile("/definitions/parameterOrReference");
+
+        if (node != null && node.getType() != null && pointer.equals(node.getType().getPointer())) {
+            // validation parameter location value
+            if (node.isObject() && node.asObject().get("in") != null) {
+                AbstractNode valueNode = node.asObject().get("in");
+                try {
+                    Object value = valueNode.asValue().getValue();
+
+                    if (!Lists.newArrayList("query", "header", "path", "cookie").contains(value)) {
```
tedepstein, line 101: @ghillairet , @tfesenko , It looks like we did this using special validation logic.  Are we generally trying to avoid making edits to the generated schema? Or have we already diverged from the generated schema? 

I'm wondering if it would be more efficient, generally, to either make these edits directly in the schema (in this case by adding an enum constraint), or post-process the generated schema to add missing details that could not be extracted by markdown-scraping.

ghillairet: @tedepstein We did that because the v3 schema does only contain a string type for this value. The v2 did contain an enum type.

tedepstein: Yes, the v2 schema was hand-crafted by someone.  The v3 schema is generated by analyzing the OpenAPI 3.0 specification's source markdown, and it misses many of these constraints. 

So my question is really about how we add those missing constraints, as we discover them. I see three general approaches:

1. Add special-case validation and content assist code, as you've done here. 
2. Directly edit the generated v3 schema (Which means we won't be able to use updates from a re-generated schema, if there is one. But that might be OK.)
3. Write some code that modifies the generated v3 schema, and run that code either at design time, or as part of our build, or at runtime. This is a little bit more difficult than  option 2, but would avoid the special-case logic of option 1, and still let us use future versions of the generated v3 schema.

Issue #379, opened 2017-08-24
Navigation from path parameters is broken in v3


ghillairet:
Clicking on a path parameter does not make the cursor move to the parameter definitions like it does in v2.
For example, in the petstore.yaml example, clicking on `petId` in `/pets/{petId}` from line 54 navigates to the path parameter

tedepstein:
Possibly related to #365?

ghillairet:
@tedepstein No it's not related. The hyperlink detectors for parameters was simply not present in OpenAPI v3 editor. I added it in PR #380

PR #380, opened 2017-08-30
[#379] Navigation from path parameters is broken in v3

ghillairet:
See #379


tfesenko:
Passed core review and QA

Issue #381, opened 2017-09-18
Introduce semantic versioning to Kaizen-OpenAPI-Editor


tfesenko:
Cross-reference to RepreZen Jira: https://modelsolv.atlassian.net/browse/ZEN-3708

PR #385, opened 2017-09-18
#381 Change version from 0.8.0 to 1.0.0

tfesenko:


tedepstein:
@tfesenko , aside from the fact that the version had to change in a lot of places, these changes look straightforward to me.  Should we merge this now? If we do, are there downstream changes we need to make in API Studio, to update the dependencies?

Issue #334, opened 2017-05-26
OpenAPI v3: validation rule for null schema types


tfesenko:
Extracted from PR #333 
We added a validation rule for null schema types in Swagger v2 specs, see PR #263 "Use of type: null masks subsequent errors"
Investigate if we need it for v3 and implement validation rules if needed.

tfesenko:
Passed QA and code review.
Thanks!

PR #391, opened 2017-09-22
[#334] OpenAPI v3: validation rule for null schema types

ghillairet:
I made a little change in the ErrorValidator to allow it to work on OpenApi3. The only remaining issue would be that the marker is not placed where is the `null` value, but on the container object. See:

screen shot 2017-09-22 at 14 24 22

screen shot 2017-09-22 at 14 24 25


com.reprezen.swagedit.openapi3.tests/src/com/reprezen/swagedit/openapi3/validation/NullValueValidationTest.xtend
```diff
@@ -0,0 +1,167 @@
...
+			                {
+			                    "nullProperty": null
+			                }
+		'''
+
+		document.set(content)
+		document.onChange()
+
+		val errors = validator.validate(document, null as URI)
+		assertEquals(0, errors.size())
```
tfesenko, line 135: Nice

tfesenko:
Passed QA and code review.
Thanks!

Issue #394, opened 2017-09-27
OAS3: Template "Schema with a Required Property" inserts at the wrong indent level


tedepstein:
![image](https://user-images.githubusercontent.com/692813/30917416-f5752d30-a369-11e7-9bd0-b5475f9096ab.png)

ghillairet:
Fixed in PR #397 

![sc](https://user-images.githubusercontent.com/148045/30963206-e9daf2d2-a44c-11e7-9a7b-31a3c98719ce.gif)

PR #397, opened 2017-09-28
[#394] OAS3: Template "Schema with a Required Property" inserts at th…

ghillairet:
…e wrong indent level

See #394


tfesenko:
Looks good!
Passed QA and code review.

Issue #227, opened 2016-10-18
Better code completion proposal for name/key of patterned properties and array elements


tfesenko:
We use the `_key_` as a code assist proposal for the "no proposal" case, usually it means property name, or some other named element, represented as `additionalProperties` or `patternProperties` in the schema:
<img width="231" alt="screen shot 2016-10-17 at 8 47 55 pm" src="https://cloud.githubusercontent.com/assets/644582/19460745/80e8e05a-94ac-11e6-900b-5634751ae4c8.png">
and
<img width="181" alt="screen shot 2016-10-17 at 8 48 04 pm" src="https://cloud.githubusercontent.com/assets/644582/19460771/b82c42e6-94ac-11e6-8e7e-b0988109d4b0.png">

It can be easily changed as it's defined in com.reprezen.swagedit.assist.SwaggerProposalProvider.createObjectProposals(ObjectTypeDefinition, AbstractNode, String):

``` java
if (proposals.isEmpty()) {
    proposals.add(new Proposal("_key_" + ":", "_key_", null, null));
}
```

Note that we can change the label, the replacement text, or both.

@tedepstein , we wonder if you have ideas how to improve it, which labels/replacement use in this case.

tedepstein:
@tfesenko , can the suggestion be context-sensitive, so we have something like `(property name)` in the context shown, and some other appropriate label in other contexts?

tedepstein:
@tfesenko , @ghillairet , I'm placing high priority on this one, though I don't necessarily think we should jump into implementation. There are some things we should probably do first, just to make sure that OpenAPI v3 support is complete. 

But this is a weak point in the usability of KaiZen editor right now.  It's also closely related to #70, and maybe these two issues should be merged.  Generally, the problem is with user-specified key names, represented in the schema as `additionalProperties` or `patternProperties`.  With additionalProperties, we propose `_key_`. With pattern properties, we propose the pattern, as it's specified in the schema. In both cases, it's confusing to the user, and if the user selects the proposal, it puts it in verbatim, with no hint to the user that they should put in a proper name, and this immediately puts the editor in an error state.

# Basic Usability Improvements

So far, we've suggested the following improvements:

* Make the proposal name more descriptive of what's expected.  For example `schema_name` instead of `_key_`.  Maybe use the title of the property subschema to determine the proposal string. 
* If the user selects the proposal, select the inserted name so the user can immediately type the intended, legal name. 

[...]

PR #402, opened 2017-10-04
#227 Better code completion proposal for name/key of additional properties

tfesenko:
This PR provides better completion proposals and code templates for properties described as additionalProperties in OAS3 and Swagger v3

# OAS3
## Code assist
`(<element> name)` is proposed for elements described with `additionalProperties` in JSON Schema. When chosen, the new name is selected for editing.
In this screencast, code assist suggests `(schema name)` for a schema and then `(property name)` for a schema describing a property:
additionalpropertiesassist2
Similar approach applies to other components (`callbacks`, `links`, `examples`, etc) and other `additionalProperties`:
screen shot 2017-10-04 at 1 23 35 pm
I am not attaching screenshots for all supported cases as I have provided tests for them.

### Tests
[`OpenApi3ContentAssistProcessorTest`](https://github.com/RepreZen/KaiZen-OpenAPI-Editor/pull/402/files#diff-850eaf3ece660d4c974d721e2c4a162b) provides tests for all major cases:
* Schema
* Schema Property 
* All other components (Links, Callbacks, Parameters, Examples, RequestBodies, Headers, SecuritySchemes)
* Media types
* Encodings
* Server variables

Note that some elements still have strange names:
* Discriminator#mappings: `_type_`
* OauthFlow#Scopes:  `_type_`
* Parameter names in Links: `(any name)`

Current status is covered by tests.
 
### Implementation
`JsonProposalProvider` creates proposal for the `additionalProperties` elements. The label is described by a new `TypeDefinition#getLabel()` which tried to get needed information from JSON Schema. For the schema property, we can't infer the needed name - `property` - from the schema, so I added a `title` property to JSON Schema.
### JSON Schema
I had to add one line to the JSON Schema to provide a correct `(property name)` instead of `(schema name)`
To do it, I updated our JSON Schema generator. It's generated using commit 83638080c70784ca84fcfbf86af948c6e98bcbba on
https://github.com/RepreZen/gnostic/tree/task/KZOE-227

## Code templates
Named schemas and properties now have code templates for major cases. These code templates are generated by `Activator.addNamedSchemaTemplatesInSchemas()` and `addNamedSchemaTemplatesInSchemaProperties()` using already existing code templates defined for inline schemas in `templates.xml`.
I also added code template for a schema ref for schema properties.
additionalpropertiestemplates
### Tests
* I added a new code-template context for schema properties to `SchemaObject.yaml`. It means that our JUnit tests verify that:
1. This context exists at a given location
2. All code templates provided for this context at this location will produce a valid OAS3 documentation.
### Known issues
Restore Defaults on KaiZen>OpenAPI v3/Templates preferences page erases synthetic code templates generated for named schema and named property.

### Implementation
* I added a new context for named schema properties
* `Activator.addNamedSchemaTemplatesInSchemas()` and `addNamedSchemaTemplatesInSchemaProperties()` build code templates dynamically
* Usually we use template variable to define template parts that need to be selected. Variable names don't allow spaces or special characters, and we need it to be something like `(schema name)`. So I added `ElementNameResolver` to handle it.
---------

# Swagger v2
## Code assist
Similar code assist experience was implemented for Swagger v2:
screen shot 2017-10-04 at 2 03 24 pm
screen shot 2017-10-04 at 2 03 39 pm
I provided several tests for `(schema name)` and `(property name)` in `SwaggerProposalProviderTest`, but I did not test it as thorough as foer OAS 3.

## Code templates
I did NOT provide synthetic code templates for named schemas in Swagger v2. Anyway, only one code template is available for a named schema,

## JSON Schema
Again, JSON Schema was modified to provide a correct value for `(property name)`, but as JSON Schema for Swagger v2 is not generated, there is no need to update the schema generator.


tfesenko:
@ghillairet , @tedepstein , this PR is ready for review.

tedepstein:
Looks good, Tanya!  I will plan to test it after it has passed code review and been merged.

com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/schema/ComplexTypeDefinition.java
```diff
@@ -139,4 +140,9 @@ public boolean validate(AbstractNode valueNode) {
 
         return isValid;
     }
+    
+    @Override
+    public String getLabel() {
+        return Iterables.getFirst(complexTypes, null).getLabel();
```
ghillairet: You will get a NPE if complexTypes is null here.

ghillairet (approved):
Passed QA, all tests passed as well.

I just left a comment about a possible NPE. 

Also I noted that there is a content assist proposal `_key_` present in  responses > content > schema. Not sure why it is there.

screen shot 2017-10-06 at 15 19 39

com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/schema/ComplexTypeDefinition.java
tfesenko: Thanks! Addressed in  9f2cd7d

tfesenko:
@ghillairet , thanks for reporting a case with a `_key_` proposal for an anonymous schema in media types. I can't reproduce it - I even created a test for it in 3a4675d.
Can you please double-check it and share the OAS3 file with me?

ghillairet:
@tfesenko See my commit that adds a test to show that _key_ is still present for response content schema.

tfesenko:
@ghillairet , I extracted #411 "`_type_` as context assist proposal for context schema" which can be addressed later.
Merging this PR as it passed your code review and QA.

Issue #310, opened 2017-05-22
OpenAPI v3: code assist for response code creates invalid value


tfesenko:
Code assist proposal for response code creates an invalid value: `([0-9X]{3})$:`.
It happens because we use a pattern regex for pattern properties.

As a workaround, the users can use code templates for a response:
<img width="409" alt="screen shot 2017-05-22 at 9 30 28 am" src="https://cloud.githubusercontent.com/assets/644582/26311368/0c7b72de-3ed2-11e7-8520-5c76eb91a607.png">

ghillairet:
@tedepstein @tfesenko I found a list of status codes in a json document, maybe we could use it to generate proposals.

That could look like this:

![screen shot 2017-09-22 at 16 35 41](https://user-images.githubusercontent.com/148045/30749618-3caefe9c-9fb4-11e7-8a43-712cd131512a.png)

tedepstein:
Nice idea, @ghillairet !  It should apply for OpenAPI 2.0 and 3.0. If it's easy (2 hours or less), please go ahead with implementation. 

(BTW, I was going to ask whether these should be simple values or templates. With templates, the advantage is that we can provide the response code (key) and the boilerplate template (value) at the same time.  Disadvantage is that we would have to create a lot of templates, with a lot of duplication.  It's probably not worth it. So I suggest just including the code as a key value, as shown in your screenshot.)

ghillairet:
Related https://github.com/RepreZen/KaiZen-OpenAPI-Editor/issues/70

ghillairet:
See PR #396 

This is how I implemented it so far:

![paths](https://user-images.githubusercontent.com/148045/30962865-b59e16d0-a44b-11e7-9128-186343a749ed.gif)

ghillairet:
Also it could be ported to v2 as well.

tfesenko:
Fixed in PR #396.
The replacementString should whap the response code into single quotes `200:` -> `'200':`. It's an easy change. I merged this PR now but keeping the issue open.

ghillairet:
@tfesenko PR for single quotes https://github.com/RepreZen/KaiZen-OpenAPI-Editor/pull/403

tfesenko:
Reopening:
* While numeric status codes (e.g. `'200'`) should be single-quoted, the `default` value should not be quoted. 
See 
```yaml
'200':
  description: a pet to be returned
  content: 
    application/json:
      schema:
[...]

PR #403, opened 2017-10-04
[#310] OpenAPI v3: code assist for response code creates invalid value

ghillairet:
Added single quotes in replace string.

See #310


tfesenko:
Thanks!

PR #409, opened 2017-10-05
[#310] OpenAPI v3: code assist for response code creates invalid value

tfesenko:
@ghillairet changes for #310
Remove single quotes for default


tfesenko:
Looks good, thanks!

PR #412, opened 2017-10-11
[#308] OpenAPI v3: Validation for the value of Link#operationId

ghillairet:
See #308 

The error message is now ` Invalid operationId. This operationId does not match any existing and resolvable operation.`


tfesenko:
Passed code review and QA.
Thanks!

Issue #410, opened 2017-10-06
References to path items are not treated correctly


tfesenko:
## Problem
com.reprezen.swagedit.core.model.Model.find(String) always returns `null` for path items.
Why is that:
* `Model.nodes` is a table which uses **un-encoded** JsonPointers as the keys, e.g. `/paths/~1test~1{id}`
* `Model.find` look-ups the `Model.nodes` using **encoded** JsonPointers for path items, e.g. `#/paths/~1test~1%7Bid%7D`

## Symptoms
The symptoms below are true for reference to path items with `{` or `}`

### 1 You can use a reference to a path item anywhere and the validator will be OK with it.
Example:
```yaml
---
openapi: "3.0.0"
info:
  version: "1.0.0"
  title: Test API
  
paths:
  /test/{id}:
    get:
      summary: Read
      description: Provide details for the entire list (for collection resources) or an item (for object resources)
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                # This is a reference to the parent path item,
                # but i should be a reference to a schema!!!
                $ref: "#/paths/~1test~1%7Bid%7D"
```

### 2 `operationRef` shows a validation error when operation path has `{` or `}`
At the same time, if you try to use a reference to child operation in a valid context as a value for `operationRef`, it creates an error about an incorrect type:
```yaml
---
openapi: "3.0.0"
info:
  version: "1.0.0"
  title: Test API
  
paths:
  /test/{id}:
    get:
      summary: Read
      description: Provide details for the entire list (for collection resources) or an item (for object resources)
      responses:
        '200':
          description: OK

components:
  links:
    myLink:
      # A valid ref creates an error
      operationRef: "#/paths/~1test~1%7Bid%7D/get"
```

PR #413, opened 2017-10-11
[#410] References to path items are not treated correctly

ghillairet:
See #410 

Model.find method now decodes pointers to handle special characters such as { and }.


tfesenko:
Passed code review and QA.

Issue #374, opened 2017-08-10
Validate format constraints on string properties as described in OpenAPI 3.0 spec


daffinm:
For example, the following fragment gives no warnings or errors and yet it is not legal according to [the specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md). 
```
---
openapi: "3.0.0"
info:
  version: "1.0.0"
  title: Whatever you like
  description: API_description
  termsOfService: EXPECT ERROR: Spec says "MUST be in the format of a URL."
  contact:
    name: contact_name
    url: EXPECT ERROR: Spec says "MUST be in the format of a URL."
    email: EXPECT ERROR: Spec says "MUST be in the format of an email address."
  license:
    name: MIT
    url: EXPECT ERROR: Spec says "MUST be in the format of a URL."
paths: {}
```
@tedepstein comments:

> It's a general problem that traces back to the fact that our JSON schema is auto-generated by a markdown spec-scraping utility built by Tim Burks for the Google Gnostic OSS project. That utility doesn't scrape the format requirement, so we have to scour the spec for these constraints. A search through the spec for "Must be in the format" should help here.

tedepstein:
@daffinm , thanks for logging this.  Please apply the "OpenAPIv3" label to these issues, as I've done on this one.

ghillairet:
PR #415 shows how to add email validation.

ghillairet:
In PR #415 I put some screencasts showing the current state of url/email validation. Those work in the v2 editor because the v2 schema includes the format `email` and `uri` for those properties. 

But as shown in the screencasts the validation of url values may be not what we expect (for example a value `foo` is consider valid URI). The JSON schema validator that we are using, constructs a java.net.URI object to validate those values (see [code](https://github.com/java-json-tools/json-schema-validator/blob/master/src/main/java/com/github/fge/jsonschema/format/common/URIAttribute.java#L62)). According to the JSON schema specification, this is correct because only the format `uri` is considered. There is a [discussion](https://github.com/json-schema-org/json-schema-spec/issues/233) about whether adding `url` as format for URL validation should be considered or not.  

The OpenAPI v3 specification states that those properties should be valid URL. If that means that in this case URI should not be considered, then we could implement our own URL validator (that uses java.net.URL instead of java.net.URI). Or we could consider the current validation sufficient and only adds the missing `format` properties in the v3 schema.

tedepstein:
[...]

PR #415, opened 2017-10-12
[#374] Validate format constraints on string properties as described …

ghillairet:
…in OpenAPI 3.0 spec

See #374 

Modified schema to validate email property

### DIFF
--- a/com.reprezen.swagedit.openapi3/src/com/reprezen/swagedit/openapi3/schema/schema_v3.json
+++ b/com.reprezen.swagedit.openapi3/src/com/reprezen/swagedit/openapi3/schema/schema_v3.json
@@ -102,7 +102,8 @@
           "type": "string"
         },
         "url": {
-          "type": "string"
+          "type": "string",
+          "format": "uri"
         },
         "email": {
           "type": "string",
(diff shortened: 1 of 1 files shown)


ghillairet:
@tfesenko See this PR, it includes changes in the schema to allow validation of email properties. I haven't found how to validate url properties yet, maybe we will need to to it ourselves.

media2

com.reprezen.swagedit.openapi3/src/com/reprezen/swagedit/openapi3/schema/schema_v3.json
```diff
@@ -105,7 +105,8 @@
           "type": "string"
         },
         "email": {
-          "type": "string"
+          "type": "string",
+          "format": "email"
```
tfesenko, line 110: 👍

ghillairet:
@tfesenko The v2 schema already defines the correct format for those properties (see https://github.com/RepreZen/KaiZen-OpenAPI-Editor/blob/master/com.reprezen.swagedit/src/com/reprezen/swagedit/schema/schema.json#L129) The email validation is working. The uri validation does not work in the editor but works when I run tests on it. I'm trying to understand why it does not show up in the editor.

ghillairet:
The validation done by the JSON schema validator already uses java URI to check the validity of the URI but the results may not be what is expected, see

media3

That can be reproduced in current v2 editor.

With changes I made to implement this functionality myself, I get the same results.

media4

tfesenko:
See https://github.com/googleapis/gnostic/pull/63

Issue #419, opened 2017-10-24
Autocomplete: case is not corrected if picked from dropdown


jfiala:
applies to swagger2 + openapi3 editor:

e.g. if you type "minlength" for a query parameter and then pick "minLength" from the list, the case is  not correct, the wrongly typed case remains as typed.

PR #422, opened 2017-10-24
#419 Autocomplete: case is not corrected if picked from dropdown

tfesenko:
#419 Autocomplete: case is not corrected if picked from dropdown


ghillairet (approved):
Passed QA

Issue #439, opened 2018-05-02
Validation: Allow references to components in non-OAS files, through any path


tedepstein:
KZOE is too restrictive in validating references.  It correctly ensures that the references themselves are in contexts explicitly allowed by OpenAPI 2.0 and 3.0. But it _incorrectly_ requires the referenced components to be reachable through an OAS-standard path. 

The OpenAPI 2.0 spec shows [examples of references](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#referenceObject) to single-component files (or "fragment files"), and to components in non-OAS-standard paths:

> ##### Reference Object Example
> 
> ```js
> {
> 	"$ref": "#/definitions/Pet"
> }
> ```
> 
> ```yaml
> $ref: '#/definitions/Pet'
> ```
> 
> ##### Relative Schema File Example
> ```js
> {
>   "$ref": "Pet.json"
> }
> ```
> 
> ```yaml
> $ref: 'Pet.yaml'
> ```
> 
> ##### Relative Files With Embedded Schema Example
> ```js
> {
>   "$ref": "definitions.json#/Pet"
> }
> ```
> 
> ```yaml
> $ref: 'definitions.yaml#/Pet'
> ```
> 

The OpenAPI 3.0.1 spec has [equivalent examples](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#referenceObject), and neither spec explicitly says that referenced components have to be in their designated locations, in a valid OpenAPI document.

So we should not impose these additional restrictions.

# Example: Reference to single-component schema file

Given a `Quotations_Schema.yaml` file like this:
```
title: Quotations
description: received result for quotations
type: object
properties:
  zing:
    type: string
  zang:
    type: string
```

... the following OpenAPI 3.0 spec in the same folder should be valid:
```
---
openapi: "3.0.0"
info:
  version: "1.0.0"
  title: BeamUp API

paths:
  /products:
    get:
      responses:  
        200:
          description: An array of products
          content:
            application/json:
              schema:
                $ref: "Quotations_Schema.yaml"
```
It fails to resolve the ref, showing a warning:

`Invalid object reference, the referenced object is not of expected type.`

Here's an OpenAPI 2.0 file that should also be valid in the same folder containing `Quotations_Schema.yaml`.  It fails with the same warning.

```
---
swagger: "2.0"
info:
  version: 1.0.0
  title: TaxBlaster

paths:
[...]

Issue #354, opened 2017-06-08
Better validation message for unexpected object type for $ref properties


tfesenko:
Extracted from #323 "Validation of $ref property values should show error on unexpected object type"
The current error message just says "Invalid object reference, the referenced object is not of expected type."  It does not provide information about the expected type:

<img width="768" alt="screen shot 2017-06-08 at 5 22 17 pm" src="https://user-images.githubusercontent.com/644582/26952443-7bd8eb8a-4c73-11e7-8794-2a20a820e009.png">

We should see which information we can get from the JSON Schema and provide it in the message

ghillairet:
One possible solution is to give the pointer of the current type, and the pointer of the expected type, like this:

![screen shot 2017-06-09 at 16 53 02](https://user-images.githubusercontent.com/148045/26981038-444792e0-4d34-11e7-8283-2e34fce1dec9.png)

Test model:

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
paths:
  /:
    get:
      responses:
        200:
          $ref: "#/components/schemas/Pet"             
components:
  responses:
    Ok:
      description: ok
  schemas:
    Pet:
      properties:
        id:
          type: number
```

tedepstein:
This is an example of where "title" could come in very handy.

ghillairet:
Unfortunately `title` is not available everywhere in the schema, for example `responseOrReference` is defined like this:

```json
"responseOrReference": {
      "oneOf": [
        {
          "$ref": "#/definitions/response"
        },
        {
          "$ref": "#/definitions/reference"
        }
      ]
    },
```

And response is defined like this:

```json
"response": {
      "type": "object",
      "description": "Describes a single response from an API Operation, including design-time, static  `links` to operations based on the response.",
      "required": [
        "description"
      ],
      "additionalProperties": false,
      "patternProperties": {
        "^x-": {
          "$ref": "#/definitions/specificationExtension"
        }
      },
      "properties": {
        "description": {
          "type": "string"
        },
        "headers": {
          "$ref": "#/definitions/headers"
        },
        "content": {
          "$ref": "#/definitions/content"
        },
        "links": {
          "$ref": "#/definitions/links"
        }
      }
    }
```

I could change the message to include only the last path of the pointer, so it would be `It should be of type responseOrReference but it is of type schemaOrReference`.

tfesenko:
[...]

PR #440, opened 2018-05-03
[#439] Validation: Allow references to components in non-OAS files, t…

ghillairet:
Validation: Allow references to components in non-OAS files, through any path

This commit modifies the ReferenceValidator to allow references to non OAS/Swagger files. The type validation of referenced elements is still performed but
now uses the JsonSchema validator library instead of our builtin mechanism based on JSON schema pointers. The new type validation is done by obtaining the subset of the OAS/Swagger JSON schema for which the reference is typed and validate the actual target node on it. If validation succeeds then the reference is marked as valid.

### DIFF
--- a/com.reprezen.swagedit.core/META-INF/MANIFEST.MF
+++ b/com.reprezen.swagedit.core/META-INF/MANIFEST.MF
@@ -22,7 +22,11 @@ Require-Bundle: org.eclipse.ui,
  org.eclipse.ui.workbench.texteditor
 Bundle-RequiredExecutionEnvironment: JavaSE-1.7
 Bundle-ActivationPolicy: lazy
-Export-Package: com.reprezen.swagedit.core,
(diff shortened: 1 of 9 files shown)


tedepstein:
@andylowry , please review when you can.

andylowry (approved):
Passed review and QA.

andylowry:
@tedepstein FYI, a small nit on this:

When the reference does resolve, but the target JSON structure does not pass the validation to which it is subjected, a single generic error message is provided: 
> Invalid object reference, the referenced object is not of expected type.

The validation itself creates its own messages, but those are not examined or displayed. It might be preferable to display the specific messages, perhaps, as markers nested beneath the main marker. If you'd prefer that, I'd suggest that we make it a follow-on issue.

tedepstein:
@andylowry , thanks.  The issue with validation messages is covered in #354.

Issue #438, opened 2018-05-02
"Unable to resolve the reference" error on http $ref URL with non-standard port


vnalla:
I have a OpenAPI 3.0.0 compliant YAML spec files which work fine in Swagger Editor. KaiZen Eclipse (Oxygen) plugin is not able to resolve references. The files are served from tomcat server on the same machine where eclipse is running. Tried with draft 4 compliant JSON Schema file as well as YAML file.

```
Invalid Reference - Unable to resolve the reference
The value must be valid JSON Reference (for external references) or JSON Pointer (for local references), and must resolve to an object of expected type
```

![image](https://user-images.githubusercontent.com/11826854/39545889-95c5c4f0-4e20-11e8-9829-d7a05223958e.png)

thanks

tedepstein:
@vnalla, I've been reviewing some existing issues, and doing some extensive testing. 

With your first reference, I think there are likely two problems:

1. `CreatAddress.yaml` looks like a typo.  Should be `CreateAddress.yaml`, I assume.  The specific error message you're getting from KZOE -- "The value must be valid JSON Reference..." -- indicates that it is not able to resolve the reference as specified.  Please double-check to make sure the specified URL is correct, and let us know if you're still getting _the same_ error message...

2. Even if your path is correct, I expect that you'll see a _different_ error message, saying "Invalid object reference, the referenced object is not of expected type."  

This is due to a problem detailed in issue #439, which I have just logged. Apparently, KZOE's validation is too restrictive.  It wants to find the referenced parameters, schemas, etc. in an _OAS-standard path_, i.e. a path that could contain an object of the expected type, in a valid OpenAPI document. 

The schema you're referencing is in a file by itself. Presumably the schema is the root object in that file. A reference to an object at root level, or in any location that isn't a standard path for an OAS object of the expected type, will fail to resolve. 

I am not 100% sure about the timeline, but I hope we'll be able to get this fixed quickly.  In the meantime, one possible workaround is to create an OpenAPI 3.0 file with an empty `paths` object, put the reusable components in their designated place, and reference them there. 

See the attached zip file for an example of a multi-file project with component files.

[Multi-file Beamup (v3).zip](https://github.com/RepreZen/KaiZen-OpenAPI-Editor/files/1969252/Multi-file.Beamup.v3.zip)

vnalla:
[...]

PR #441, opened 2018-05-03
[#438] Does not resolve local or external references

ghillairet:
Missing URL port during normalization of URLs.

### DIFF
--- a/com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/json/references/JsonDocumentManager.java
+++ b/com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/json/references/JsonDocumentManager.java
@@ -112,9 +112,12 @@ public class JsonDocumentManager {
         }
     }
 
+    /*
+     * Returns a copy of the URL parameter without the fragment or query.
+     */
     private URL normalize(URL url) {
         try {
-            return new URL(url.getProtocol(), url.getHost(), url.getFile());
+            return new URL(url.getProtocol(), url.getHost(), url.getPort(), url.getFile());
         } catch (MalformedURLException e) {
             return url;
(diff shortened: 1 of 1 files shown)


com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/json/references/JsonDocumentManager.java
```diff
@@ -112,9 +112,12 @@ private JsonNode parse(URL url) {
         }
     }
 
+    /*
+     * Returns a copy of the URL parameter without the fragment or query.
+     */
```
andylowry, line 117: Misleading comment. Only the fragment is omitted. `URL#getFile()` returns the path portion plus the query portion, if it exists. Not using the query would be an error, since query strings are often required to properly address documents.

andylowry: Might also be worth converting to URI and normalizing the path, and using that as a key for loaded docs. `URL#equals` does hostname resolution, so using it as a map key will do so as well. It's therefore not generally recommended as a map key or set element.

The result isn't perfect. URL.equals does its host resolution to recognize when two host names really refer to the same host, and URI doesn't do that. Also, rearrangements of query params are generally meaningless, so it'd be nice to sort them or something; but that's not guaranteed, so you can't do that. Unfortunately, URL/URI identity detection is not easy. But requiring hostname resolution in map operations is probably best avoided.

andylowry:
@ghillairet 

@tedepstein wanted me to cut a new build today for this so the issue reported can check it out. I'll leave it to you to do what you will with my review comments.

PR #456, opened 2018-08-14
ZEN-4176 Prevent validation from running multiple times

tfesenko:
The "Update KaiZen Editor validation markers" job now cancels other
validations on the current file if a new validation is initiated.

In general, validation is invoked:
* Inside a bigger WorkspaceJob on save(), saveAs(), and validation
preferences change for Swagger v2
* As an individual WorkspaceJob "Update KaiZen Editor validation
markers" on document opening (`setInput()`) and
change(`documentChanged()`)

Extracted a lightweight `ValidationOperation` which:
* Cancellable
* Provides more accurate progress report


ghillairet (approved):
Looks good, passed code review and QA.

PR #458, opened 2018-08-20
ZEN-4351 Invalid Thread Access Exception thrown when saving document and other fixes

tfesenko:
Improvements:
* Don't run validation in UI thread as it worsens the user experience
* Make JsonDocument's `model`, `jsonContent`, and `yamlContent` multi-thread safe
* Fix potential NPE by only passing non-null Model to `JsonReferenceValidator`

# ZEN-4351 Invalid Thread Access Exception thrown when saving documents
## Monitor.isCancelled() and the Main Thread.
Normally, monitor.isCancelled() should NOT be executed in the Main
thread - that's the idea of the progress monitor.

However, the save operations use a special implementation of the
progress monitor (`EventLoopProgressMonitor`)  which is 
> used to run an event loop whenever progress monitor methods are
invoked

That's why `isCancelled()` operation must be executed in the Main Thread
for this implementation of the monitor. 

`org.eclipse.ui.internal.SaveableHelper.saveModels()` creates and
`EventLoopProgressMonitor`, then it calls `doSaveModel()` which calls
the editor's save method (`model.doSave(subMonitor.split(2),
shellProvider);`). This is how our `doSave()` methods gets this
implementation of the monitor and passes it to the validation.

### DIFF
--- a/com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/editor/JsonDocument.java
+++ b/com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/editor/JsonDocument.java
@@ -12,6 +12,7 @@ package com.reprezen.swagedit.core.editor;
 
 import java.io.IOException;
 import java.io.StringReader;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.jface.text.BadLocationException;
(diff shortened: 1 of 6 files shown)


ghillairet (approved):
Passed code review and QA.

PR #460, opened 2018-08-23
ZEN-4329 Review and improve code templates for OpenAPI v3

tfesenko:
We now have three code templates for resources
* Collection Resource: GET and POST
* Object Resource: GET, PATCH, and DELETE
* Singleton Resource: GET, PUT, and DELETE 

GET has three variations:
* Generic version
* For a collection resource: with an `x-next` header and returning an array of items
* Object inside a collection resource: with a path parameter returning a single item

Added code templates for commonly used schemas:
* Error 
* Links

We can also consider adding these elements:
* `Range` header 
* ETag


tfesenko:
With @tedepstein's permission, merging to master.

Issue #459, opened 2018-08-22
Latest version update throwing exception


vnalla:
![image](https://user-images.githubusercontent.com/11826854/44473234-d8f45400-a5fd-11e8-8568-69bf5b8ef44c.png)

tfesenko:
Hi @vnalla , Sorry to hear that you are having this issue with KaiZen-OpenAPI-Editor and thank you for reporting it. 
I can't see this problem on my build. Can you please provide more details so I could reproduce it:
* Version of KaiZen-OpenAPI-Editor
* Eclipse distribution which you use as a host for KaiZen-OpenAPI-Editor
* How to reproduce this error. E.g., is it happening on opening an OAS3 file? 
* Do you have anything in the Error Log?

These details will help to diagnose the issue. Thanks!

vnalla:
Version of KaiZen-OpenAPI-Editor -- 0.8.0.201808221157
Eclipse Version: Orxygen.3a Release (4.7.3a) Build id:20180405-1200
Yes for OAS3 files
[ErrorLog.log](https://github.com/RepreZen/KaiZen-OpenAPI-Editor/files/2311424/ErrorLog.log)

vnalla:
I also tried to install in the Eclipse Photon version as well, but it does not have OAS3 Spec.

![image](https://user-images.githubusercontent.com/11826854/44480352-03034180-a611-11e8-9a70-5386f57028a7.png)

Eclipse Version: Photon Release (4.8.0) Build id: 20180619-1200
KaiZen-OpenAPI-Editor version: 0.8.0.201808221157 -- latest version from Eclipse Market place
[ErrorLog1.log](https://github.com/RepreZen/KaiZen-OpenAPI-Editor/files/2311447/ErrorLog1.log)

tfesenko:
@ghillairet , @andylowry , this is our latest build (from six hours ago), the problem is in duplicate instances of `javax.mail`.
```
org.osgi.framework.BundleException: Could not resolve module: com.reprezen.swagedit.openapi3 [1098]
  Unresolved requirement: Require-Bundle: com.github.eclipsecolortheme; resolution:="optional"
  Bundle was not resolved because of a uses contraint violation.
[...]

PR #461, opened 2018-08-23
#459 - remove dependency on javax.mail

tfesenko:
Note: I wasn't able to reproduce the issue reported in #459 on my Mac. The issue was initially found on Windows.

`javax.mail` is only used in the `validateEmailField()` method which is used by `ContactValidator`.
Validation showed correct results on a Contact object when the `email` value was correct and when it was incorrect. 

I tested the update site created with this change and didn't have any problems in validation of `Contact` or other elements.

### DIFF
--- a/com.reprezen.swagedit.openapi3/META-INF/MANIFEST.MF
+++ b/com.reprezen.swagedit.openapi3/META-INF/MANIFEST.MF
@@ -24,8 +24,7 @@ Require-Bundle: org.eclipse.ui,
  org.apache.log4j;bundle-version="1.2.15",
  com.reprezen.jsonoverlay;bundle-version="3.1.0",
  com.reprezen.kaizen.openapi-parser;bundle-version="3.0.1",
- com.norconex.commons.norconex-commons-lang;bundle-version="1.15.0",
- javax.mail-api;bundle-version="1.6.1"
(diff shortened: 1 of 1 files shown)


com.reprezen.swagedit.openapi3/META-INF/MANIFEST.MF
```diff
@@ -24,8 +24,7 @@ Require-Bundle: org.eclipse.ui,
  org.apache.log4j;bundle-version="1.2.15",
  com.reprezen.jsonoverlay;bundle-version="3.1.0",
  com.reprezen.kaizen.openapi-parser;bundle-version="3.0.1",
- com.norconex.commons.norconex-commons-lang;bundle-version="1.15.0",
- javax.mail-api;bundle-version="1.6.1"
+ com.norconex.commons.norconex-commons-lang;bundle-version="1.15.0"
```
andylowry, line 27: Didn't know you were using norconnex for this. I'm using it in JsonOverlay to canonicalize URLs, to maximize likelihood of using shared parse results when two different-looking references actually refer to the same document.

Since I'm already using it, I'll look into it for validations like email as well.

andylowry (approved):
Looks good, and checks out in QA

PR #466, opened 2018-08-26
ZEN-4360, ZEN-4365 - Editor background color set to white by default, doesn't follow Color Theme

tfesenko:


tfesenko:
This PR fixes the color theme. By doing this, it also improves the appearance of the closing curly so it's now visible:
screen shot 2018-08-26 at 6 15 45 pm

tfesenko:
@andylowry , I am merging this PR now. Please review it when you have a moment.

com.reprezen.swagedit.openapi3/resources/com.reprezen.swagedit.openapi3.xml
```diff
@@ -13,5 +13,7 @@
 		<mapping pluginKey="colorTagPropertyPreferences" themeKey="foreground" />
 		<mapping pluginKey="colorFlowCharacterPreferences" themeKey="foreground" />
 		<mapping pluginKey="colorConstantPreferences" themeKey="method" />
+        <mapping pluginKey="AbstractTextEditor.Color.Foreground" themeKey="foreground"/>
```
tedepstein, line 16: @tfesenko , small thing, but it looks like the indent level is not consistent. Maybe a tabs vs. spaces problem.

com.reprezen.swagedit/resources/com.reprezen.swagedit.xml
```diff
@@ -13,5 +13,7 @@
 		<mapping pluginKey="colorTagPropertyPreferences" themeKey="foreground" />
 		<mapping pluginKey="colorFlowCharacterPreferences" themeKey="foreground" />
 		<mapping pluginKey="colorConstantPreferences" themeKey="method" />
+        <mapping pluginKey="AbstractTextEditor.Color.Foreground" themeKey="foreground"/>
```
tedepstein, line 16: @tfesenko , same problem with indents here. 

More importantly, I'm inferring from this that, for some reason, the highlighting logic in effect here is not highlighting certain characters, under certain circumstances. So those characters are defaulting to the `PREFERENCE_COLOR_FOREGROUND` key defined in [`AbstractTextEditor`](https://help.eclipse.org/oxygen/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fui%2Ftexteditor%2FAbstractTextEditor.html). And that color wasn't mapped to Eclipse Color Theme, so it defaults to black. Is that right? 

The screenshot shows that it's improved, but I take it this is not the final solution. We need to figure out why the highlighter is not doing the right thing.

PR #467, opened 2018-08-27
ZEN-4361 Missing marker location indicators (Overview Ruler) next to editor scrollbar in KZOE

tfesenko:
Preferences store for EditorsPlugin has settings to show/hide the rules
and markers. 
We were using EditorsPlugin's preferences store before commit
e63addb99483563426d75c4e4b6ecfc98cf47d96 when we replaced it by our
editor's pref store. This commit uses a composite (chained) preferences
store which has read access to both EditorsPlugin and KZOE editor's
editor stores and writes values to the KZOE editor's prefs.


tfesenko:
@ghillairet I merged this PR, but please review it when you have the time.

tedepstein:
Yay, @tfesenko !!!  🥇

ghillairet (commented):
Looks good

PR #468, opened 2018-08-27
ZEN-4368 - Consider PATCH instead of PUT in OAS3 Object Resource code template

tfesenko:


com.reprezen.swagedit.openapi3/resources/templates.xml
```diff
@@ -73,7 +73,7 @@ email: ${contact_email}</template>
               $$ref: '${itemSchema}' 
       '404':
         description: Not Found
-  put:
+  patch:
     description: Create a new ${item}
```
tedepstein, line 77: The description should be `Modify an existing ${item}`

Issue #469, opened 2018-08-28
"Object resource" template fails to put braces around path parameter in path


andylowry:
You end up with a path string of e.g. `pets/petId` instead of `/pets/{petId}`, and this causes the overall template to contain error markers where `petId` parameters are declared for operations.

PR #470, opened 2018-08-28
[#469] Fix Object resource template

andylowry:
This realates to issue ZEN-4367 RepreZen JIRA

The `Object resource` template does not work well with reference string
content-assist when filling in its `itemSchema` placeholder. The first
instance of the placeholder is fine, but the next one has an extra quote
in front of the ref string. This appears to be a problem with the
content-assist implementation, not something that can be addressed by
changing the template. So this commit does not fix the reported problem.

However, while investigating this issue, another bug in that template
came to light, namely that the path parameter specified via the "id"
plaaceholder in the template ends up without needed braces in the path
string. This causes corresponding parameter definitions in contained
operations to be in error.

This commit fixes the path string issue, but as to the reference string
issue, it only makes quoting of ref strings consistent throughout the
templates. Where ref strings are quoted, they are now always quoted with
double quotes, whereas before there appeared to be a somewhat random use
of double or single quotes.

### DIFF
--- a/com.reprezen.swagedit.openapi3/resources/templates.xml
+++ b/com.reprezen.swagedit.openapi3/resources/templates.xml
@@ -54,7 +54,7 @@ email: ${contact_email}</template>
        id="com.reprezen.swagedit.openapi3.templates.object_resource_template" 
        description="object resource template" 
        context="com.reprezen.swagedit.openapi3.templates.paths" 
-       enabled="true">/${item}s/${id}:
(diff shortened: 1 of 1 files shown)


tfesenko (approved):

tedepstein:
@andylowry , these changes look worthwhile. I added some comments on the issue, and included a hack, in the form of a comment block with instructions on how to complete the unquoted schema reference using code assist. Maybe you can incorporate that here.

PR #471, opened 2018-08-28
[#469] Remove quotes from some ref placeholders (WIP)

tfesenko:
Where the same reference appears more than once in a single template,
the placeholders are no longer quoted. This is due to observed buggy
behavior by eclipse in filling templates.


tfesenko:
@andylowry , I think you fixed all the occurences. 
Merging

PR #472, opened 2018-08-29
ZEN-4357 Evaluate Guava dependencies, consider removing where practical

tfesenko:
Guava is a recurring problem with Eclipse. We should consider removing our use of Guava, instead using FP and other features of Java 8.

I removed usages of Guava in KZOE, but I still need to keep a dependency on Guava in our manifest declaration because [JSON Schema Validator uses Guava](https://github.com/java-json-tools/json-schema-core/blob/master/src/main/java/com/github/fge/jsonschema/core/load/configuration/URIDownloadersRegistry.java)


andylowry (approved):
Whew! Looks good.

com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/assist/JsonContentAssistProcessor.java
```diff
@@ -196,7 +195,7 @@ protected void clearStatus() {
             int offset) {
         final List<ICompletionProposal> result = new ArrayList<>();
 
-        prefix = Strings.emptyToNull(prefix);
+        prefix = "".equals(prefix) ? null : prefix;
```
ghillairet: We should be using `StringUtils.emptyToNull` here.

ghillairet (approved):
@tfesenko Looks good. I made 2 commits to solve an issue with Arrays.asList(). This method creates an immutable list, and we were using it to create lists that are then updated. 

It also removes guava from tests MANIFEST and target platform as it should not be needed as Guava comes with Xtend.

com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/assist/JsonContentAssistProcessor.java
ghillairet: @tfesenko I did the change.

ghillairet:
@tfesenko Please review my 2 commits and merge if you think it's ok.

com.reprezen.swagedit.core/src/com/reprezen/swagedit/core/editor/outline/OutlineContentProvider.java
```diff
@@ -48,7 +48,7 @@ public void inputChanged(Viewer viewer, Object oldInput, Object newInput) {
             return null;
         }
 
-        List<Object> roots = Arrays.asList();
+        List<Object> roots = new ArrayList<>();
```
tfesenko: Thanks!

PR #473, opened 2018-09-02
ZEN-4367, ZEN-3845  - Resource templates don't populate schema reference correctly

tfesenko:
### ZEN-3845: Code assist should replace selected text
KZOE doesn't seem to have an option to overwrite the selected text. The CTRL key doesn't seem to have any effect.

I would even go further and say that the idea of an "insert" mode is quite counterintuitive. If I have text selected when I invoke code assist, I would fully expect that whatever I select from the suggestion list will replace the selected text. If it doesn't, it seems like a bug. I would never think to hold down the CTRL key, even if it worked. And I would not think to look for an option in preferences to make code assist replace instead of insert. I would just assume it's a bug. Apparently Andy Lowry make the same assumption in ZEN-4367 .

I think "replace mode" should be at least the default mode, and probably the only mode. It's not worth any extra effort to provide an option key (like Ctrl) or a preference setting for "insert mode" in code assist; I don't see the point.

### ZEN-4367:  Resource templates don't populate schema reference correctly
I tried inserting and filling in the collection resource code template, and found that there were warnings due to invalid references.
The first $ref variable is between single quotes; the second is within double quotes. After using code assist to choose a valid reference, the second one looks like this:
`$ref: "'#/components/schemas/TaxFiling"`
There is a stray single quote after the opening double quote.
The Object Resource template has a similar problem.

## Changes Overview
The major changes are done in `StyledCompletionProposal` and `ProposalDescriptor`
* `Proposal` -> `ProposalDescriptor` with fluent syntax to set all the values, added `toString()`
* Simplified creation of `StyledCompletionProposal` instances
* `StyledCompletionProposal`: added new `preSelectedRegionLength` to address ZEN-3845 Code assist should replace selected text

## Tests
`JsonReferenceProposalProvider_Quotes_Test` provides integration tests for code assist for elements with quotes (references), it also checks the document state after the code assist proposal have been applied.


tfesenko:
@ghillairet , this PR is ready for your review

com.reprezen.swagedit.openapi3.tests/src/com/reprezen/swagedit/openapi3/assist/JsonReferenceProposalProvider_Quotes_Test.xtend
```diff
@@ -0,0 +1,224 @@
...
+
+		override protected getContextTypeRegistry() { null }
+
+		override protected getTemplateStore() { null }
+
+		override protected getContextTypeId(Model model, String path) { null }
+
+	}
+
+	/*============    Code below copied from utils.Cursors, otherwise Xtend won't compile     ============= */
```
ghillairet, line 171: @tfesenko Any clue why it does not compile for you? I have it working.

ghillairet (changes requested):
@tfesenko Looks good. It works as expected, although I would prefer we remove the duplicate in the tests.

tfesenko:
> although I would prefer we remove the duplicate in the tests.

Covered by https://modelsolv.atlassian.net/browse/ZEN-4428, will be fixed later.

Merging.

PR #480, opened 2019-01-16
[ZEN-4529] KZOE: Quick outline view with filter duplicates outline branches

ghillairet:
The quick outline tree showed duplicates because equality of nodes was not implemented correctly, so the tree could not filter the nodes.

Before:
screen shot 2019-01-16 at 12 23 59

After:
screen shot 2019-01-16 at 12 22 30


tedepstein:
@tedepstein to QA and code review.

tedepstein:
Passed QA, fix is very straightforward.

tedepstein:
Passed QA. Merging.

PR #481, opened 2019-01-17
[ZEN-4499] KZOE Editor: Typeahead doesn't filter code templates correctly

ghillairet:
This PR adds filtering of template proposals during typeahead. The filter returns the template proposals for which the display string contains the current typeahead string.

Screencasts:

Swagger:
s1

OpenAPI
s2

s3


tedepstein:
Passed QA. Nice!

PR #483, opened 2019-02-06
[ZEN-4744] KZOE: Security requirement without security schemes borks validation

ghillairet:
This PR fixes a NPE that was happening when security schemes were not present inside components. The validation now doesn't silently fails and error markers are correctly set.

s3


tedepstein:
@ghillairet , the change looks fine and straightforward. I made a small tweak to the error message, directly on the task branch.  It built without errors, so I will go ahead and merge. If you have any concerns about that, please let me know and we can reopen this.

PR #487, opened 2019-03-28
[ZEN-3764] Example Data: Generate example dummy data into OpenAPI Model from single schema (DO NOT MERGE)

nbhusare:
This PR add's support for providing example data in the KZOE editor, in the form of a proposal provider. It introduces a new extension point named **com.reprezen.swagedit.core.exampleprovider**. Plug-ins having a dependency on **com.reprezen.swagedit.core** can provide an extension to the aforementioned extension point. The extension receives a resolved schema instance as input, which can be used with libraries like Adobe Ride (model-util), JSON Schema Faker... to generate dummy data.

It also enhances the internal utility API by adding two new classes - ModelUtils and ExtensionUtils. ModelUtils has API for traversing the JSON Node model (AST) and retrieving nodes with a given name. The ExtensionUtils adds support for reading extension points. 
 
`JsonExampleProposalProvider` is the base class of the newly introduced example providers for Swagger and OpenAPI 3.0. The implementations `OpenApi3ExampleProposalProvider` and `SwaggerExampleProposalProvider` provide intelligence of identifying the content assist location and retrieving the ExampleProvider implementation.


com.reprezen.swagedit.openapi3/build.properties
```diff
@@ -5,4 +5,5 @@ bin.includes = META-INF/,\
                OSGI-INF/,\
                plugin.xml,\
                icons/,\
-               resources/
+               resources/,\
+               schema/
```
ghillairet: @nbhusare You should remove this `schema` it is making the build to fail.

PR #493, opened 2019-04-11
[ZEN-4782] Example Validation: Use error severity for schema validation failures

ghillairet:
Update preferences to allow users to select the level of validation for examples.


tedepstein:
To do by @ghillairet : Only enable validation failure severity if example validation is checked.

tedepstein:
@neerajbhusare , this has passed QA. 

There's a comment above about enabling failure severity only if example validation is checked, otherwise disabling...  But this is captured in another Jira issue, so it can be dealt with separately.

So please review the code and merge if you're OK with it.

nbhusare:
Code looks good.

PR #494, opened 2019-04-11
[ZEN-4800] OAS v2 document template contains invalid example.

ghillairet:


tedepstein (approved):
I had attached a document to the Jira issue with somewhat different corrections. But this is fine. Merging.

PR #497, opened 2019-04-24
 [ZEN-4802] Add Quick Fixes on example validation markers (MERGE AFTER #493)

ghillairet:


tedepstein:
As discussed, we want to make the Quick Fix actions more discoverable, so you don't need to use the QuickFix Hot key.  

* **Most valuable** - Clicking the error marker should show the quick fix.  
2019-04-24 07_35_09-Demo_Prasanthi - foo_src_com_reprezen_testproj_Testing java - C__Users_tedep_Rep
* **Lower priority** - JDT also displays a dotted underline on ranges of code that caused the error; and displays quick fixes on hover. This is not as important, and it's not worth spending a lot of time. If we can implement it without too much effort, it does have some value.  If it is problematic to apply the underline to the specific part of the example that does not validate, we could apply the underline to the contents of the line of code that has the error marker.

tedepstein:
@neerajbhusare , if this has passed your QA and code review, please merge it. 

In this scope, we won't be providing the Quick Fixes through mouse interactions (click on marker, hover over code). I've opened a separate issue to track that work.

tedepstein:
@neerajbhusare , BTW, we created https://github.com/ModelSolv/RepreZen/pull/1887 to test this, and have already merged that one. The normal process would be to merge the component PR (this one) first, update dependencies API Studio PR, then merge the API Studio PR.

Issue #496, opened 2019-04-22
Code assist is missing properties named "default"


tedepstein:
Raised in [this comment](https://marketplace.eclipse.org/comment/6190#comment-6190) on Eclipse Marketplace.

In OpenAPI 3.0, the following objects have properties named `default`:
* Schema Object
* Server Variable Object
* Responses Object

Of these three, KaiZen Editor only offers `default` as a content assist proposal in the context of the Responses Object.  Schema Object and Server Variable object should also include `default` in the proposal list, but they currently do not. 

In OpenAPI 2.0, the following objects have `default` properties:
* Parameter Object (non-body parameters)
* Items Object
* Responses Object
* Header Object

Again, code assist only suggests `default` in the context of the Responses Object. It should be present in all of these contexts.

PR #218, opened 2016-10-05
Content assist enhancements 

ghillairet:
This PR fixes the following issues relative to content assist
- #216 - Content-assist: strange description for the values of `format`
- #213 - Code-assist does not work for `in` for top-level parameters
- #214 - Code-assist: `required` show twice for path method parameters

PR #498, opened 2019-04-25
[#496 / ZEN-4808] Code assist is missing properties named "default"

ghillairet:
Addresses #496 .


tedepstein:
Passed QA.  I tested this on OAS2 (Swagger) and OAS3 documents. I confirmed that `default` now shows up as expected in code assist proposals; and confirmed that there were no regressions w.r.t. the specific issues that PR #218 was intended to address.   Merging.

PR #503, opened 2019-05-21
Update target platform to be based on Oxygen

ghillairet:
This also removes the project containing the kepler target platform and the associated profile as they are not used anymore.


tedepstein:
@ghillairet , I was reviewing this with @neerajbhusare today, and we are not sure which issues this addresses. We are assuming for now that this is _not_ required for the Confluence GenTemplate beta release. We can look at it after that release and merge it in. 

If it needs to be included now, please let us know ASAP.

PR #505, opened 2019-06-18
[ZEN-4921] Parameter and header code assist is incomplete; style parameter should show enumerated values

ghillairet:
Added style values as enum in openapi schema.


com.reprezen.swagedit.openapi3/src/com/reprezen/swagedit/openapi3/schema/schema_v3.json
```diff
@@ -416,7 +416,16 @@
...
+          "type": "string",
+          "enum": [
+            "matrix",
+            "label",
+            "form",
+            "simple",
+            "spaceDelimited",
+            "pipeDelimited",
+            "deepObject"
+          ]
```
tedepstein, line 428: The `style` property also applies to `header` (see [Header Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#headerObject)) and `encoding` ([Encoding Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#encodingObject)). I think it makes sense to extract this string enum type as its own `styleEnum` schema, and reference that schema from all three contexts.

PR #518, opened 2019-08-15
[ZEN-4990] Remove requirement for a KZOP update site

ghillairet:
We strip the versions of the dependencies that are copied into the lib folder so that build scripts don't have to modify the MANIFEST file.
This file should still be updated and committed if any change requires it. For example if a dependency is added or removed.

The maven build will also add a file named classPath.txt into the lib folder. This is done by the maven dependency plugin. This file will contain dependency versions that are needed by the gentemplate wizard to set versions of gentemplate projects dependencies (such as genflow, kzop, rapidml).


com.reprezen.swagedit.dependencies/META-INF/MANIFEST.MF
```diff
@@ -95,6 +96,6 @@ Export-Package: com.fasterxml.jackson.annotation,
  org.yaml.snakeyaml.serializer,
  org.yaml.snakeyaml.tokens,
  org.yaml.snakeyaml.util
-Require-Bundle: com.google.guava;bundle-version="21.0.0",
+Require-Bundle: com.google.guava;bundle-version="[14.0.0,22.0.0)",
```
tedepstein, line 99: Can KZOE really work with this range of Guava releases? 

I know we have done some work to eliminate guava dependencies from our components. How much work remains to completely eliminate the Guava dependency from KZOE?

com.reprezen.swagedit.dependencies/pom.xml
```diff
@@ -53,22 +78,71 @@
...
+							<stripVersion>true</stripVersion>
+							<excludeArtifactIds>com.reprezen.swagedit.dependencies</excludeArtifactIds>
+							<excludeGroupIds>
+								com.google.guava,
+								javax.mail,
+								javax.activation,
+								com.sun.mail
+							</excludeGroupIds>
+							<excludeScope>system</excludeScope>
+							<includeScope>runtime</includeScope>
```
tedepstein, line 96: I see that this `<configuration>` was pushed down from `<plugin>` level to `<execution>` level, presumably because you added a new execution with its own separate configuration. 

Also, the configuration new has `<excludeScope>` and `<includeScope>` properties. Can you explain these?

com.reprezen.swagedit.dependencies/META-INF/MANIFEST.MF
```diff
@@ -6,25 +6,26 @@ Bundle-ActivationPolicy: lazy
...
+ lib/com.github.java-json-tools.json-schema-core.jar,
+ lib/com.github.java-json-tools.json-schema-validator.jar,
+ lib/com.google.code.findbugs.jsr305.jar,
+ lib/com.googlecode.libphonenumber.libphonenumber.jar,
+ lib/com.reprezen.jsonoverlay.jsonoverlay.jar,
+ lib/com.reprezen.kaizen.openapi-parser.jar,
+ lib/joda-time.joda-time.jar,
+ lib/net.sf.jopt-simple.jopt-simple.jar,
+ lib/org.mozilla.rhino.jar,
+ lib/org.yaml.snakeyaml.jar
```
tedepstein, line 28: Why don't we need the version suffixes anymore?

com.reprezen.swagedit.dependencies/build.properties
```diff
@@ -1,20 +1,2 @@
...
-               lib/com.github.java-json-tools.json-schema-validator-2.2.10.jar,\
-               lib/com.google.code.findbugs.jsr305-3.0.1.jar,\
-               lib/com.googlecode.libphonenumber.libphonenumber-8.0.0.jar,\
-               lib/com.reprezen.jsonoverlay.jsonoverlay-4.0.4.jar,\
-               lib/com.reprezen.kaizen.openapi-parser-4.0.4.jar,\
-               lib/joda-time.joda-time-2.9.7.jar,\
-               lib/net.sf.jopt-simple.jopt-simple-5.0.3.jar,\
-               lib/org.mozilla.rhino-1.7.7.1.jar,\
-               lib/org.yaml.snakeyaml-1.23.jar
+               lib/
```
tedepstein, line 2: It looks like we previously listed `/lib` _AND_ then listed individual .jar files under /lib.  Now we only include `/lib`. Why...?

com.reprezen.swagedit.dependencies/pom.xml
```diff
@@ -32,8 +32,33 @@
 		</dependency>
 	</dependencies>
 
+	<repositories>
+		<repository>
+			<id>sonatype-snapshots</id>
```
tedepstein, line 37: It looks like we are adding the sonatype repo so that dev builds of KZOE can pick up snapshot releases of KZOP. And this is needed now, I assume, because we're no longer using an update site for KZOP. Is that right?

com.reprezen.swagedit.dependencies/META-INF/MANIFEST.MF
ghillairet: It's not about KZOE, it's the schema validator that uses Guava. We need to have a version that is compatible with the Guava version used by Xtend. And this is the range that uses Xtend. See https://github.com/ModelSolv/RepreZen/pull/1871

com.reprezen.swagedit.dependencies/pom.xml
```diff
@@ -53,22 +78,71 @@
...
+					</execution>
+					<execution>
+						<id>classpath</id>
+						<goals>
+							<goal>build-classpath</goal>
+						</goals>
+						<configuration>
+							<prefix>lib</prefix>
+							<prependGroupId>true</prependGroupId>
+							<outputFile>${project.basedir}/lib/classPath.txt</outputFile>
```
tedepstein, line 107: It looks to me like the purpose of this configuration is to create the `${project.basedir}/lib/classPath.txt`  file, but I don't see that filename referenced anywhere else. Is it used by one of our build scripts on TC?

com.reprezen.swagedit.dependencies/pom.xml
ghillairet: The `<excludeScope>` and `<includeScope>` are there to prevent eclipse dependencies pulled by tycho to be considered.

com.reprezen.swagedit.dependencies/META-INF/MANIFEST.MF
ghillairet: It is explained here https://github.com/RepreZen/KaiZen-OpenAPI-Editor/pull/518/commits/aa52a267670b8c1cb35b8727b67b6f29b3e0fcec

Shortly build scripts could not update those so we should have versionless names.

com.reprezen.swagedit.dependencies/pom.xml
ghillairet: yes

com.reprezen.swagedit.dependencies/pom.xml
ghillairet: Because it's produced by maven during project build.

com.reprezen.swagedit.dependencies/pom.xml
```diff
@@ -53,22 +78,71 @@
...
+										<groupId>org.apache.maven.plugins</groupId>
+										<artifactId>maven-dependency-plugin</artifactId>
+										<versionRange>[2.8,)</versionRange>
+										<goals>
+											<goal>build-classpath</goal>
+										</goals>
+									</pluginExecutionFilter>
+									<action>
+										<execute>
+											<runOnIncremental>false</runOnIncremental>
```
tedepstein, line 137: OK, this whole `<pluginManagement>` construct was added to prevent execution of the `build-classpath`  goal each time there's an incremental build in Eclipse, right?

ghillairet: That's specific to the maven eclipse plugin. It's to make sure the `build-classpath` goal will be executed by maven inside eclipse.  Otherwise it may not do it and you will have to build the project or execute that goal manually to see it executed in eclipse.

com.reprezen.swagedit.dependencies/pom.xml
tedepstein: OK, I think the explanation is here, in the same commit comment you referenced above:

> The maven build will also add a file named classPath.txt into the lib folder. This is done by the maven dependency plugin. This file will contain dependency versions that are needed by the gentemplate wizard to set versions of gentemplate projects dependencies (such as genflow, kzop, rapidml).

So the Maven build produces it, and it's used at runtime by the GenTemplate wizard.

tedepstein:
@ghillairet , thanks for the explanations. I can merge now.

